PR #1, opened 2023-08-14
chore: add append-only log

arriqaaq:
This PR adds an append-only log (which can also be used as a write-ahead log)

TODO:

- [ ] Add support for compression
- [ ] Create an lru cache for all the segments to keep them open for reads and to avoid creating unnecessary fds per read
- [ ] Add tests
- [ ] Add support for detecting corrupted records
- [ ] Add support to fix corrupted log
- [ ] Use custom error than io::Result
- [ ] Check to support Direct I/O


src/storage/aol/aol.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub fn append(&mut self, rec: &[u8]) -> io::Result<(u64, usize)> {
+        if self.closed {
+            return Err(io::Error::new(io::ErrorKind::Other, "Segment is closed"));
+        }
+
+        if rec.is_empty() {
+            return Err(io::Error::new(io::ErrorKind::Other, "buf is empty"));
+        }
+
+        let _lock = self.mutex.write().unwrap();
```
emmanuel-keller, line 113: Regarding the locks from sync (RwLock and Mutex), in a real-world situation, you have to handle the possible poisoning of the lock. You will have this error if a previous lock has panicked.

https://doc.rust-lang.org/std/sync/struct.PoisonError.html
  
```Rust
 match shared_db.read() {
        Ok(read_guard) => {
            println!("Read data: {:?}", *read_guard);
        },
        Err(poisoned) => {
            println!("Detected poison during read! Recovering...");
            let read_guard = poisoned.into_inner();
            println!("Read data: {:?}", *read_guard);
        }
    }

    // And let's try a write operation after the panic
    match shared_db.write() {
        Ok(mut write_guard) => {
            write_guard.insert(4, "Daisy".to_string());
        },
        Err(poisoned) => {
            println!("Detected poison during write! Recovering...");
            let mut write_guard = poisoned.into_inner();
            write_guard.insert(4, "Daisy".to_string());
        }
    }
```

Alternatively, you may have a lock at `parking_lot`, which is simple to use and very often used in replacement.
https://crates.io/crates/parking_lot

src/storage/aol/aol.rs
```diff
@@ -0,0 +1,354 @@
...
+
+        // Get options and initialize variables
+        let opts = &self.opts;
+        let mut n = 0usize;
+        let mut available = 0;
+        let mut offset = 0;
+
+        while n < rec.len() {
+            // Calculate available space in the active segment
+            available = opts.max_file_size.unwrap() - self.active_segment.offset();
```
emmanuel-keller, line 123: Is it okay to panic if there is no value in the Option? Or maybe you want to use a default value using `unwrap_or` ?

```
opts.max_file_size.unwrap_or(0)
```

src/storage/aol/aol.rs
```diff
@@ -0,0 +1,354 @@
...
+        assert_eq!(&[4, 5, 6, 7, 8, 9, 10].to_vec(), &bs[8..]);
+
+        let mut bs = vec![0; 15];
+        let r = a.read_at(&mut bs, 4097);
+        assert!(r.is_err());
+
+        // Cleanup: Drop the temp directory, which deletes its contents
+        drop(temp_dir);
+    }
+}
```
emmanuel-keller, line 354: Adding a test with concurrency to check the locks?

src/storage/aol/segment.rs
```diff
@@ -0,0 +1,532 @@
...
+        }
+
+        Ok(n)
+    }
+}
+
+impl Drop for Segment {
+    /// Attempt to fsync data on drop, in case we're running without sync.
+    fn drop(&mut self) {
+        self.sync().ok();
```
emmanuel-keller, line 409: May we want to log that? So we have a chance to know that at some point in the code, `sync` has not been called before the segment has been dropped?

emmanuel-keller (approved):
Excellent!

src/storage/aol/segment.rs
arriqaaq: Ack, I am adding that

mumoshu (approved):

Miuler:
This is not `chore:` is a `feat:`

> chore: updating grunt tasks etc; no production code change

Please stop using chore, it is not suitable for these commits that modify the code.

- https://medium.com/neudesic-innovation/conventional-commits-a-better-way-78d6785c2e08
- https://stackoverflow.com/questions/26944762/when-to-use-chore-as-type-of-commit-message
- https://github.com/angular/angular/commit/dff6ee32725197bdb81f3f63c5bd9805f2ed22bb#diff-6a3371457528722a734f3c51d9238c13

PR #5, opened 2023-11-20
Add KV store

arriqaaq:
This is a draft PR for the KV store with a bunch of pending changes as listed below:

## Store Changes
- [x] Fix Range scans (scans don't stop at end bound)
- [x] Remove dependency from VariableKey during scan reads
- [x] Add hermitage tests for SSI
- [x] Add Per KV checksum
- [x] Fix options which are not implemented
- [x] Add hermitage tests for SI
- [x] Make BLOCK_SIZE = 32kb
- [x] Add key reader (scan support) to track reading multiple keys in a transaction
- [x] Add Read Cache
- [x] Store all database options in file header
- [x] Add close method to store
- [x] Fix value reference decode
- [x] Need to terminate keys when inserting to snapshot and index (Foo vs foo2 bug (bytes need to be appended with delimiter “\0” as in vector key for art to differentiate))
- [x] Improve Store API (create vector/array key from byte array instead of making the user import Vector/Array key, and remove dependency on bytes::Bytes for txn api)


src/storage/cache/queue.rs
```diff
@@ -0,0 +1,203 @@
+use core::marker::PhantomData;
+use std::{
+    ptr::{null_mut, NonNull},
+    sync::atomic::{AtomicPtr, Ordering::*},
+};
+
+pub struct Queue<T> {
```
DelSkayn, line 7: This seems to be a reimplementation of `std::collections::LinkedList`. Is there any reason why you couldn't use `LinkedList`?

src/storage/cache/queue.rs
```diff
@@ -0,0 +1,203 @@
...
+                Some(val) => {
+                    unsafe {
+                        let next = head_ptr.as_ref().next.load(Acquire);
+                        if !next.is_null() {
+                            self.head.swap(next, Relaxed);
+                        }
+                    };
+                    self.size -= 1;
+                    return Some(val);
+                }
```
DelSkayn, line 52: Memory leak, the `Node` object which contained the value taken after `unsafe { head_ptr.as_mut().item.take() }`  is still allocated.

src/storage/kv/transaction.rs
```diff
@@ -0,0 +1,1163 @@
...
+
+        {
+            let txn1 = store.begin().unwrap();
+            let mut txn2 = store.begin().unwrap();
+
+            assert!(txn1.get(&key3).is_err());
+            txn2.set(&key2, &value3).unwrap();
+            txn2.commit().unwrap();
+
+            assert!(txn1.get(&key3).is_err());
```
mumoshu, line 947: (Sorry if I misunderstood something badly, but) I think we want to avoid this anomaly either (1)by failing a read when it is about to return the data that causes a pmp anomaly OR (2)by returning the data from the snapshot associated to the tx.

Which behavior are you trying to check with this test?

src/storage/kv/transaction.rs
```diff
@@ -0,0 +1,1163 @@
...
+
+        let key2 = Bytes::from("k2");
+        let key3 = Bytes::from("k3");
+        let value3 = Bytes::from("v3");
+
+        {
+            let txn1 = store.begin().unwrap();
+            let mut txn2 = store.begin().unwrap();
+
+            assert!(txn1.get(&key3).is_err());
```
mumoshu, line 943: Shouldn't this and the same op in L947 be a scan against `v1..=v3`?
It seems to me that currently two txn1's read sets and the txn2's write set don't overlap so there might be no chance of an anomaly even if it wasn't based on MVCC?

mumoshu: Also, what kind of error are we currently expecting here? (A missing key error, by any chance?

For a PMP test, I was expecting we would check that two txn1 scans before and after the key insertion made by the txn2 give the same result sets.

But again, I might be misunderstanding something badly! Maybe you could give me a brief explanation of what you are trying to check with this test, so that I can figure it out myself.

src/storage/cache/queue.rs
arriqaaq: Thanks for the suggestion @DelSkayn, I want to make this a lockfree queue, and wanted to experiment with lower level rust. I wasn't aware of the LinkedList, but will give it a go.

src/storage/kv/transaction.rs
arriqaaq: Thanks for the comment @mumoshu 

In a transactional context, pmp here is refering to a situation to test if multiple transactions are well isolated from each other, and read predicates are isolated.

Given the two transactions, what I am trying to check is that even though txn2 is committed, txn1 still has its own snapshot and won't read data which has been committed post it began.

Does that answer your question? Please feel free to lemme know. Ideally this fails in read committed or other isolation levels. Example [here](https://github.com/ept/hermitage/blob/master/sqlserver.md)

src/storage/kv/transaction.rs
arriqaaq: Ah, my bad. The error is KeyNotFound. 

1) I will add the explicit error assertion to make it more clear.
2) With regards to scan, I did not use it because there is no predicate search as such yet in the KV store. The tests in hermitage usually are doing a select based on value, so I wasn't sure of adding a scan to read all the keys. 

This is the [reference](https://github.com/ept/hermitage/blob/master/foundationdb.md)

If you think scans will make it more explicit, let me know. I wanted to search for a specific key (searching for key3) to make it more readable.

src/storage/cache/queue.rs
arriqaaq: I need to fix this, thanks for catching this and mentioning it before too.

src/storage/index/snapshot.rs
```diff
@@ -0,0 +1,282 @@
...
+use crate::storage::index::node::Version;
+use crate::storage::index::KeyTrait;
+
+/// Represents a snapshot of the data within the Trie.
+pub struct Snapshot<P: KeyTrait, V: Clone> {
+    pub(crate) id: u64,
+    pub(crate) ts: u64,
+    pub(crate) root: Option<Rc<Node<P, V>>>,
+    pub(crate) readers: HashSet<u64>,
+    pub(crate) max_active_readers: AtomicU64,
```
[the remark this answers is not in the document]
arriqaaq: Thanks for the pointer, changed this to Cell

src/storage/index/snapshot.rs
```diff
@@ -0,0 +1,282 @@
...
+        Ok(())
+    }
+
+    /// Closes the snapshot, preventing further modifications, and releases associated resources.
+    pub fn close(&mut self) -> Result<(), TrieError> {
+        // Check if the snapshot is already closed
+        self.is_closed()?;
+
+        // Check if there are any active readers for the snapshot
+        if self.max_active_readers.load(Ordering::SeqCst) > 0 {
```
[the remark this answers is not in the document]
arriqaaq: Noted, changed this a Cell so this won't be used

src/storage/index/node.rs
```diff
@@ -0,0 +1,1134 @@
...
+// in that array index a separate array of up to 48 pointers.
+//
+// A Node48 is a 256-entry array of pointers to children. The pointers are stored in
+// a Vector Array, which is a Vector of length WIDTH (48) that stores the pointers.
+
+pub struct Node48<P: KeyTrait + Clone, N: Version> {
+    pub(crate) prefix: P,
+    pub(crate) version: u64,
+    child_ptr_indexes: Box<SparseVector<u8, 256>>,
+    children: Box<SparseVector<Rc<N>, 48>>,
```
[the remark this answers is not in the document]
arriqaaq: As per the discussion with you, I'd like to work on this in a separate PR. Will note these issues.

src/storage/cache/queue.rs
finnbear: I think there are lock-free queues in `crossbeam`, possibly including these:
- https://docs.rs/crossbeam/latest/crossbeam/queue/struct.SegQueue.html
- https://docs.rs/crossbeam/latest/crossbeam/deque/index.html

src/storage/cache/s3fifo.rs
```diff
@@ -0,0 +1,215 @@
...
+        let mut evicted = false;
+        while !evicted && self.main.len() > 0 {
+            if let Some(tail) = self.main.pop() {
+                let freq = tail.freq.load(Relaxed);
+                if freq > 0 {
+                    tail.freq.store(freq - 1, Relaxed);
+                    self.main.push(tail);
+                } else {
+                    self.table.remove(&tail.key);
+                    evicted = true;
```
finnbear, line 125: Consider `break;` instead

src/storage/index/iter.rs
```diff
@@ -0,0 +1,289 @@
...
+    type Item = (Vec<u8>, &'a V, &'a u64, &'a u64);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while let Some(node) = self.forward.iters.last_mut() {
+            let e = node.next();
+            match e {
+                Some(other) => {
+                    if other.1.is_twig() {
+                        let NodeType::Twig(twig) = &other.1.node_type else {
+                            panic!("should not happen");
```
finnbear, line 256: `unreachable!()` is more idiomatic than `panic!("should not happen");`

src/storage/kv/transaction.rs
mumoshu: > what I am trying to check is that even though txn2 is committed, txn1 still has its own snapshot and won't read data which has been committed post it began.

Thanks! But txn2 writes to key2 only and tx1 reads from key3 == no intersection of reads and writes. So, this test won't fail even if it had such anomaly in reality, no?

More specifically, this test would fail only when the datastore confuses writes to key2 as writes to key3, which is something worse than the anomaly we'd like to detect here...?

src/storage/kv/transaction.rs
mumoshu: Thanks! Re point 2, I think you'd rather test for imp rather than pmp then.

It looks like to me that pmp is a specialization(?) of imp; if you take the predicate used here as "match all (a predicate returns true for anything)", used `scan` instead of `get` for testing it, it would look much more like what is described in `pmp` test described in https://github.com/ept/hermitage/blob/master/foundationdb.md and https://sitano.github.io/theory/databases/2019/07/30/tx-isolation-anomalies/#imppmp-predicate-many-preceders

Currently, this "test_snapshot_isolation_pmp" looks like testing something like the below if written in a hermitage-like way:

```
    tell 0 "begin"
    tell 1 "begin"
    tell 0 "select * from test where key = 'key3'" # Returns nothing
    tell 1 "insert into test (key, value) values('key2', 'value3')" # You write to another key
    tell 1 "commit"
    tell 0 "select * from test where key = 'key3'" # Still returns nothing
    tell 0 "commit"
```

The readset and writeset doesn't intersect so this would pass even when the implementation doesn't actually prevent pmp and even imp.

If we are to test for imp for now as we don't have predicates in surrealkv, a hermitage-style test would look like:


```
    tell 0 "begin"
    tell 1 "begin"
    tell 0 "select * from test" # Returns nothing
    tell 1 "insert into test (key, value) values('key2', 'value3')" # You write to another key
    tell 1 "commit"
    tell 0 "select * from test" # Still returns nothing
    tell 0 "commit"
```

Notice that I just omitted `where key = 'key3'`. Doing so would correct using `scan` in this test? 🤔

mumoshu: Or.. it could look a bit more like PMP if txn1 and txn2 reads and writes from/to the same key, assuming the predicate is a equality check against the key of the kv pairs.

Maybe use `scan` in imp test and change the txn1 and txn2 to readwrite the same key for pmp test?

Cargo.toml
```diff
@@ -7,12 +7,30 @@ edition = "2021"
...
 futures-lite = "1.12.0"
 once_cell = "1.17.1"
 flume = "0.11.0"
+bytes = "1.5.0"
 
 [dev-dependencies]
 tempdir = "0.3"
+rand = "0.8.5"
+criterion = "0.5.1"
+jemallocator = "0.3.2"
```
emmanuel-keller, line 28: jemallocator 0.3.2 seems pretty old (4 years ago),  currently 0.5.4

src/storage/index/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+
+    pub fn from_str(s: &str) -> Self {
+        assert!(s.len() < SIZE, "data length is greater than array length");
+        let mut arr = [0; SIZE];
+        arr[..s.len()].copy_from_slice(s.as_bytes());
+        Self {
+            content: arr,
+            len: s.len() + 1,
+        }
+    }
```
emmanuel-keller, line 105: This can be confused with the standard trait `std::str::FromStr::from_str`. You may either rename the method, or implement the trait.

src/storage/cache/queue.rs
arriqaaq: Thanks for the suggestion @finnbear, currently I am not using this anywhere but as suggested, will look into this

src/storage/index/iter.rs
arriqaaq: Fixed this! Thank you

Cargo.toml
arriqaaq: Thanks for pointing it out!

src/storage/index/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+// Keys can be of two types:
+// 1. Fixed-length datatypes such as 128-bit integers, or strings of exactly 64-bytes,
+// don’t have any problem because there can, by construction, never be any key that’s
+// a prefix of any other.
+//
+// 2. Variable-length datatypes such as general strings, can be transformed into types
+// where no key is the prefix of any other by a simple trick: append the NULL byte to every key.
+// The NULL byte, as it does in C-style strings, indicates that this is the end of the key, and
+// no characters can come after it. Therefore no string with a null-byte can be a prefix of any other,
+// because no string can have any characters after the NULL byte!
```
DelSkayn, line 48: Just a note that rust strings can freely contain null bytes. `"foo\0bar"` is allowed in rust and will have a lenght of 7. So in rust adding a null byte you might still end up with a prefixed string.

Maybe instead of a null byte you could add something like `0b10111111`? All rust strings are valid UTF-8 and a UTF-8 character can never begin with a `10` bit pattern as this the start pattern of a continuation byte.

arriqaaq: This is interesting! Thanks for pointing it out @DelSkayn, looking into it

DelSkayn: Just a suggestion, just ensuring that strings have no internal null bytes is also a valid solution.

src/storage/kv/transaction.rs
arriqaaq: Thanks for this @mumoshu, I have made the test as you have suggested. Do you think renaming it to IMP would be better?

src/storage/cache/queue.rs
```diff
@@ -0,0 +1,203 @@
...
+            size: 0,
+        }
+    }
+
+    #[inline(always)]
+    pub fn push(&mut self, item: T) {
+        let node = Node::new(Some(item));
+        let node_ptr = Box::into_raw(Box::new(node));
+        let prev_tail = self.tail.load(Relaxed);
+        unsafe {
```
rushmorem, line 29: One of the best practices in Rust is to add a comment explaining why `unsafe` blocks are safe in the particular context.

src/storage/cache/s3fifo.rs
```diff
@@ -0,0 +1,213 @@
...
+    ghost: Queue<K>,
+    table: HashMap<K, Entry<K, V>>,
+}
+
+impl<K, V> Cache<K, V>
+where
+    K: PartialEq + Eq + Hash + Clone + Debug,
+    V: Clone,
+{
+    pub fn new(max_cache_size: usize) -> Self {
```
rushmorem, line 49: Have you considered taking a `NonZeroUsize` to avoid having to assert below?

src/storage/index/art.rs
```diff
@@ -0,0 +1,2114 @@
...
+    where
+        R: RangeBounds<P> + 'a,
+    {
+        // If the Trie is empty, return an empty Range iterator
+        if self.root.is_none() {
+            return Range::empty(range);
+        }
+
+        let root = self.root.as_ref();
+        return Range::new(root, range);
```
rushmorem, line 1311:
```suggestion
        Range::new(root, range)
```

src/storage/index/iter.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    pub fn range<'a, R>(
+        &'a self,
+        range: R,
+    ) -> impl Iterator<Item = (Vec<u8>, &'a V, &'a u64, &'a u64)>
+    where
+        R: RangeBounds<P> + 'a,
+    {
+        return Range::new(Some(&self.root), range);
```
rushmorem, line 44:
```suggestion
        Range::new(Some(&self.root), range)
```

src/storage/index/iter.rs
```diff
@@ -0,0 +1,276 @@
...
+            Self {
+                inner: Box::new(IterState::new(node)),
+                _marker: Default::default(),
+            }
+        } else {
+            Self {
+                inner: Box::new(std::iter::empty()),
+                _marker: Default::default(),
+            }
+        }
```
rushmorem, line 102: I think a `match` statement is a bit easier to read in this case.

```suggestion
        match node {
            Some(node) => Self {
                inner: Box::new(IterState::new(node)),
                _marker: Default::default(),
            },
            None => Self {
                inner: Box::new(std::iter::empty()),
                _marker: Default::default(),
            },
        }
```

src/storage/kv/transaction.rs
```diff
@@ -0,0 +1,1336 @@
...
+        let mut index = self.store.indexer.write();
+        let mut kv_pairs = self.build_kv_pairs(tx_id, commit_ts);
+
+        index.bulk_insert(&mut kv_pairs)?;
+        Ok(())
+    }
+
+    /// Builds key-value pairs from the write set.
+    fn build_kv_pairs(&self, tx_id: u64, commit_ts: u64) -> Vec<KV<VariableKey, Bytes>> {
+        let mut kv_pairs: Vec<KV<VariableKey, Bytes>> = Vec::new();
```
rushmorem, line 389: Is the type hint necessary here? Unless I'm missing something, the compiler should be able to infer it.

```suggestion
        let mut kv_pairs = Vec::new();
```

src/storage/log/aof/log.rs
```diff
@@ -1,23 +1,25 @@
...
 /// Append-Only Log (AOL) is a data structure used to sequentially store records
 /// in a series of segments. It provides efficient write operations,
 /// making it suitable for use cases like storing large amounts of data and
-/// writing data in a sequential manner. This is useful for applications which
-/// need to store large amounts of data separately in a variety of files, and
-/// store only the offsets of the files in the main data structure. Useful for
-/// WISCKEY implementation where we need to store the key-value pairs in a
-/// separate file and store only the offsets in the main data structure.
+/// writing data in a sequential manner.
 pub struct AOL {
```
rushmorem, line 20: This should be `Aol` as per [Rust's naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html):

> In `UpperCamelCase`, acronyms and contractions of compound words count as one word: use `Uuid` rather than `UUID`, `Usize` rather than `USize` or `Stdin` rather than `StdIn`.

src/storage/log/aof/log.rs
```diff
@@ -205,31 +231,44 @@ impl AOL {
...
+            let reader = cache.get(&segment_id);
+            if reader.is_none() {
+                let segment = Segment::open(&self.dir, segment_id, &self.opts)?;
+                let read_bytes = segment.read_at(buf, read_offset)?;
+                cache.push(segment_id, segment);
+                Ok(read_bytes)
+            } else {
+                let segment = reader.unwrap();
+                segment.read_at(buf, read_offset)
+            }
```
rushmorem, line 251:
```suggestion
            match cache.get(&segment_id) {
                Some(segment) => segment.read_at(buf, read_offset),
                None => {
                    let segment = Segment::open(&self.dir, segment_id, &self.opts)?;
                    let read_bytes = segment.read_at(buf, read_offset)?;
                    cache.push(segment_id, segment);
                    Ok(read_bytes)
                }
            }
```

src/storage/log/aof/log.rs
```diff
@@ -205,31 +231,44 @@ impl AOL {
...
         Ok(())
     }
 
-    pub fn sync(&mut self) -> io::Result<()> {
-        let _lock = self.mutex.write().unwrap();
-        self.active_segment.sync()?;
-        Ok(())
+    // Returns the current offset within the segment.
+    pub fn offset(&self) -> Result<u64> {
+        let _lock = self.mutex.read();
```
rushmorem, line 263: Do we need to hold a read lock while reading `active_segment`? If so, shouldn't `active_segment` itself be protected by the mutex? Someone could just access it directly otherwise.

src/storage/log/aof/log.rs
```diff
@@ -205,31 +231,44 @@ impl AOL {
...
+        Ok(self.active_segment.offset())
     }
 
-    // Returns the current offset within the segment.
-    pub fn offset(&self) -> u64 {
-        let _lock = self.mutex.read().unwrap();
-        self.active_segment.offset()
+    pub fn size(&self) -> Result<u64> {
+        let _lock = self.mutex.read();
```
rushmorem, line 268: Same here.

src/storage/log/mod.rs
```diff
@@ -0,0 +1,2388 @@
...
+        buf.put_uint(KEY_MAX_FILE_SIZE, opts.max_file_size);
+        if let Some(md) = opts.metadata.as_ref() {
+            buf.put(KEY_ADDITIONAL_METADATA, &md.bytes());
+        }
+
+        buf
+    }
+
+    // Returns the serialized bytes representation of Metadata
+    pub(crate) fn bytes(&self) -> Vec<u8> {
```
rushmorem, line 402: I know this is not part of public API but this should be named `to_bytes` as per the naming guidelines mentioned earlier.

src/storage/log/mod.rs
```diff
@@ -0,0 +1,2388 @@
...
+        let mut b = Vec::new();
+        self.write_to(&mut b).unwrap();
```
rushmorem, line 404: Is this `unwrap` guaranteed never to panic? Would it make sense to return a result and propagate the error here?

src/storage/cache/s3fifo.rs
arriqaaq: Thanks for the suggestion; I have not. I just have one problem with the NonZeroUsize, and that is having to import it from a separate std crate. Doesn't seem friendly as an API to import a different type just to delegate the assert. But if you say this is a very common pattern in Rust and most people use it, I'd be happy to use that here.

src/storage/log/mod.rs
```diff
@@ -0,0 +1,2388 @@
...
+enum RecordType {
+    Empty = 0,  // Rest of block is empty.
+    Full = 1,   // Full record.
+    First = 2,  // First fragment of a record.
+    Middle = 3, // Middle fragments of a record.
+    Last = 4,   // Final fragment of a record.
+}
+
+impl RecordType {
+    fn from_u8(value: u8) -> Result<Self> {
```
rushmorem, line 509: Have you considered implementing `TryFrom<u8> for RecordType` instead?

src/storage/log/wal/log.rs
```diff
@@ -3,18 +3,21 @@ use std::io;
...
+use crate::storage::log::wal::reader::Reader;
+use crate::storage::log::{
+    get_segment_range, Error, IOError, MultiSegmentReader, Options, Result, Segment, SegmentRef,
+    WAL_RECORD_HEADER_SIZE,
+};
 
 /// Write-Ahead Log (WAL) is a data structure used to sequentially store records
 /// in a series of segments. It provides efficient write operations,
 /// making it suitable for use cases like write-ahead logging.
 pub struct WAL {
```
rushmorem, line 18: This should be named `Wal`.

src/storage/cache/s3fifo.rs
rushmorem: Using `NonZeroUsize` improves the type safety and it clearly communicates to the caller that this function takes only non-zero arguments without even needing to read the documentation or code. I love self documenting APIs like that. I know a `usize` looks better aesthetically but I will pick the extra steps of constructing a `NonZeroUsize` over a potential runtime panic anytime.

A `usize` would be fine if the function gave zero a special meaning, not when it panics. We could also document the panic but I don't think that's the best user experience. Panics should be avoided whenever possible. In this case, it's easy to avoid it.

src/storage/cache/queue.rs
arriqaaq: Noted, thanks @rushmorem, I have resorted to using the LinkedList from the std lib, so removing the usage of unsafe here.

src/storage/cache/s3fifo.rs
arriqaaq: I agree

src/storage/index/mod.rs
arriqaaq: noted, thank you for the information!

src/storage/kv/transaction.rs
```diff
@@ -0,0 +1,1336 @@
...
+        Ok(results)
+    }
+
+    // precommit the transaction to WAL
+    pub fn precommit(&mut self) -> Result<()> {
+        if self.store.opts.wal_disabled {
+            return Ok(());
+        }
+
+        todo!();
```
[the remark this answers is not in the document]
arriqaaq: Yes! Just implementing it in a separate PR to merge into this

src/storage/kv/transaction.rs
arriqaaq: You are right, not necessary at all

src/storage/log/aof/log.rs
arriqaaq: Noted! I wasn't aware, thank you!

src/storage/log/aof/log.rs
arriqaaq: So I've been using a mutex inside the struct to manage this, the expectation is for the end user to not use these internal fields at all, but rather the methods. The read lock just ensures the writer waits because these offsets are being used to track the position of the record in the file and don't want this to be a dirty read as such.

src/storage/log/mod.rs
arriqaaq: Thanks for suggesting this, I haven't, just looking into it.

src/storage/index/mod.rs
arriqaaq: Thanks @DelSkayn, I am referencing this issue [here](https://github.com/surrealdb/tart/issues/13) and will explore working on it in the library, as we plan to import that as the index eventually.

src/storage/index/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+impl<const SIZE: usize> Ord for FixedKey<SIZE> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.content[..self.len].cmp(&other.content[..other.len])
+    }
+}
+
+impl<const SIZE: usize> FixedKey<SIZE> {
+    // Create new instance with data ending in zero byte
+    pub fn create_key(src: &[u8]) -> Self {
+        assert!(src.len() < SIZE);
```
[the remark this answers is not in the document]
arriqaaq: Thanks @rushmorem for the suggestion, I've fixed this

src/storage/log/mod.rs
arriqaaq: I've sort of wrapped the Result here like this `pub type Result<T> = std::result::Result<T, Error>;`, and using tryfrom requires conversion between the error types

src/storage/log/mod.rs
arriqaaq: Thanks for catching this, fixed it!

src/storage/cache/queue.rs
arriqaaq: Used std crate linkedlist as you'd suggested for this

src/storage/index/node.rs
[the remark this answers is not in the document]
arriqaaq: I've removed the usage of vectors here!

src/storage/log/mod.rs
rushmorem: Isn't `Error` public in this case? You should be fine if you use the same error type in your TryFrom implementation. It's fine though. I'm not hung up on this. Using TryFrom would publicly expose this implementation anyway if RecordType is public.

src/storage/log/aof/log.rs
rushmorem: > the expectation is for the end user to not use these internal fields at all, but rather the methods.

I wasn't worried about end users. My worry was that we could access the offsets from another part of the codebase directly. Could be simply a case of forgetting to acquire the mutex first or ignorance. Over time more people are probably going to contribute to this library, not just you, which increases the risk of this happening. Rust mutexes are designed to avoid this from happening by holding the data they are protecting so that the only way you can get to it is by acquiring the mutex first.

rushmorem (approved):

src/storage/log/aof/log.rs
arriqaaq: I see your point, I'll submit a separate PR for this as I wasn't really sure on what pattern to use for accessing locks across multiple fields in Rust. Appreciate it

src/storage/kv/transaction.rs
arriqaaq: I'll keep a PR open for this if you think renaming this to IMP is right Yusuke!

PR #10, opened 2024-02-01
Make store async

arriqaaq:
### Description:

This PR introduces asynchronous functionality to the Store. The primary changes include:

- Make transaction commit async
- Refactored the Store to support async operations.
- Make test cases async

The main design change is that the store now pushes the writes to an async channel, and a background thread is consuming it and committing to the append-only log.

Looking into the usage of async aware lock(s).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ Temporary Items
 /docker/
 /.direnv/
 /testdata/
+/example/
 
 # -----------------------------------
 # Specific
--- a/.gitignore
+++ b/.gitignore
@@ -41,8 +41,6 @@ Temporary Items
 /bin/
 /docker/
 /.direnv/
-/testdata/
-/example/
 
 # -----------------------------------
 # Specific
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ comfy-table = "7.0.1"
 
 tempdir = "0.3"
 rand = "0.8.5"
-criterion = "0.5.1"
+criterion = { version = "0.5.1", features = ["async_tokio", "html_reports"] }
 jemallocator = "0.5.4"
 nanoid = "0.4.0"
 caches = "0.2" 
@@ -54,7 +54,3 @@ fastrand = "2.0.1"
 [[bench]]
 name = "store_bench"
 harness = false
-
-[[bin]]
-name = "rocks_bench"
-path = "example/rocks_bench.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ hashbrown = "0.14.2"
 lru = "0.12.0"
 async-channel = "2.1.1"
 futures = "0.3.30"
-async-std = "1.12.0"
 bytes = "1.5.0"
 tokio = { version = "1.36", features = ["rt"] }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,35 +20,12 @@ bytes = "1.5.0"
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
-
-# below is for the async runtime
-async-task = "4.4.0"
-futures-lite = "1.12.0"
-once_cell = "1.17.1"
-flume = "0.11.0"
-
-tempfile = "3.5.0"
-# for backwards compatibility testing - pin at 1.0.0
-redb = "1.5.0"
-
-# Just benchmarking dependencies
-ctrlc = "3.2.3"
-lmdb-rkv = "0.14.0"
-sanakirja = "1.3.3"
-sled = "0.34.7"
-rocksdb = "0.21.0"
-libc = "0.2.99"
-comfy-table = "7.0.1"
-
-
 tempdir = "0.3"
 rand = "0.8.5"
 criterion = { version = "0.5.1", features = ["async_tokio", "html_reports"] }
 jemallocator = "0.5.4"
 nanoid = "0.4.0"
-caches = "0.2" 
 fastrand = "2.0.1"
-
 
 
 [[bench]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,15 +13,34 @@ crossbeam-channel = "0.5.8"
 parking_lot = "0.12.1"
 hashbrown = "0.14.2"
 lru = "0.12.0"
+async-channel = "2.1.1"
+futures = "0.3.30"
+async-std = "1.12.0"
+bytes = "1.5.0"
+
+[dev-dependencies]
+tokio = { version = "1", features = ["full"] }
 
 # below is for the async runtime
 async-task = "4.4.0"
 futures-lite = "1.12.0"
 once_cell = "1.17.1"
 flume = "0.11.0"
-bytes = "1.5.0"
 
-[dev-dependencies]
+tempfile = "3.5.0"
+# for backwards compatibility testing - pin at 1.0.0
+redb = "1.5.0"
+
+# Just benchmarking dependencies
+ctrlc = "3.2.3"
+lmdb-rkv = "0.14.0"
+sanakirja = "1.3.3"
+sled = "0.34.7"
+rocksdb = "0.21.0"
+libc = "0.2.99"
+comfy-table = "7.0.1"
+
+
 tempdir = "0.3"
 rand = "0.8.5"
 criterion = "0.5.1"
@@ -30,6 +49,12 @@ nanoid = "0.4.0"
 caches = "0.2" 
 fastrand = "2.0.1"
 
+
+
 [[bench]]
-name = "s3fifo_bench"
+name = "store_bench"
 harness = false
+
+[[bin]]
+name = "rocks_bench"
+path = "example/rocks_bench.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ async-channel = "2.1.1"
 futures = "0.3.30"
 bytes = "1.5.0"
 tokio = { version = "1.36", features = ["rt"] }
+sha2 = "0.10.8"
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ futures = "0.3.30"
 bytes = "1.5.0"
 tokio = { version = "1.36", features = ["rt"] }
 sha2 = "0.10.8"
+quick_cache = "0.4.0"
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ async-channel = "2.1.1"
 futures = "0.3.30"
 async-std = "1.12.0"
 bytes = "1.5.0"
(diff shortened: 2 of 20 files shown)


benches/store_bench.rs
```diff
@@ -13,74 +17,94 @@ fn create_temp_directory() -> TempDir {
...
             .to_be_bytes()
             .into_iter()
             .cycle()
             .take(len)
-            .clone()
             .collect()
     };
 
     let mut bench = |key_len, val_len| {
+        let rt = tokio::runtime::Builder::new_multi_thread()
```
phughk, line 32: Probably doesnt need to be addressed in this PR, but was wondering if this doesnt usually get injected via the harness. Given its a benchmark then probably better to keep full control over it. But sharing the idea for consideration.

benches/store_bench.rs
```diff
@@ -126,4 +155,5 @@ fn concurrent_insert(c: &mut Criterion) {
 
 criterion_group!(benches_sequential, bulk_insert, sequential_insert_read);
 criterion_group!(benches_concurrent, concurrent_insert);
-criterion_main!(benches_sequential, benches_concurrent);
+// criterion_main!(benches_sequential, benches_concurrent);
```
phughk, line 158:
```suggestion
```

src/storage/kv/store.rs
```diff
@@ -68,22 +84,82 @@ impl Store {
...
+    ) -> Result<()> {
+        let mut txn = self.begin_with_mode(Mode::ReadWrite).await?;
         f(&mut txn)?;
-        txn.commit()?;
+        txn.commit().await?;
 
         Ok(())
     }
+
+    /// Closes the store. It sends a stop signal to the writer and waits for the done signal.
```
phughk, line 98: Im not a storage guy, but does my thinking here work?

- storage must always be synchronous and safe via the WAL
- all WAL writes are synchronous and flushed
- all page writes can be asynchronous
- all page writes must be recoverable
- recovery of a single page is as simple as swapping pointers

So this close method is optional to avoid recovery cost. Do we think it's worth keeping? It is likely useful for tests etc where there are frequent starts/stops. But was wondering

src/storage/kv/store.rs
```diff
@@ -68,22 +84,82 @@ impl Store {
...
+                }
+            }
+        }))
+    }
+
+    #[inline]
+    async fn handle_task(&self, task: Task) {
+        let core = self.core.clone();
+        if let Err(err) = core.write_request(task).await {
+            println!("failed to write: {:?}", err);
```
phughk, line 162:
```suggestion
            eprintln!("failed to write: {:?}", err);
```

phughk (approved):
Awesome :D Nice one

src/storage/kv/store.rs
```diff
@@ -25,41 +28,54 @@ use crate::storage::{
...
+            is_closed: AtomicBool::new(false),
+            task_runner_handle: Arc::new(RwLock::new(Some(task_runner_handle))),
+        })
     }
 
     /// Begins a new read-write transaction.
     /// It creates a new transaction with the core and read-write mode, and sets the read timestamp from the oracle.
     /// It returns the transaction.
-    pub fn begin(&self) -> Result<Transaction> {
+    pub async fn begin(&self) -> Result<Transaction> {
```
mumoshu, line 61: Can we perhaps keep this non-async, as none of the functions we call here are async?

src/storage/kv/store.rs
```diff
@@ -25,41 +28,54 @@ use crate::storage::{
...
         let txn = Transaction::new(self.core.clone(), mode)?;
         Ok(txn)
     }
 
     /// Executes a function in a read-only transaction.
     /// It begins a new read-only transaction and executes the function with the transaction.
     /// It returns the result of the function.
-    pub fn view(&self, f: impl FnOnce(&mut Transaction) -> Result<()>) -> Result<()> {
-        let mut txn = self.begin_with_mode(Mode::ReadOnly)?;
+    pub async fn view(&self, f: impl FnOnce(&mut Transaction) -> Result<()>) -> Result<()> {
```
mumoshu, line 77: Same as begin and begin_with_mode, perhaps we don't need to make this async if I'm not missing anything?

My understanding was that we might want to do commits asynchronously so that there might be only one thread dedicated for the commit pipeline, which can make it efficient thanks to cache locality, and option to potentially better io thanks to batching WAL writes (in exchange of durability)

And we might have no motivation to make begin and view functions async? 🤔

benches/store_bench.rs
arriqaaq: Thanks for the info @phughk, I wasn't aware, I'll look into how this is done

src/storage/kv/store.rs
arriqaaq: You are absolutely a storage guy with this thinking :)

So currently I am not committing to a WAL. The process is as follows:

1) Once a transaction is conflict free, it is selected for commit
2) The entries in this transaction are passed to a write channel as a `Task`
3) There is a thread which keeps receiving these tasks and commits them to first disk and then index
4) The Transaction is waiting on this task to get done to mark it as complete

This close method was just to ensure anything in the channel is not abruptly closed, whether I write to WAL or not.

Committing to WAL will help in case of restarts, to replay those transactions. It is for more durability guarantees. The WAL is already there, I'll add it as an optional step during precommit in a separate PR. The reason I haven't added it yet is because I wanted to take care of compaction on the WAL and the transaction log together. 

WDYT?

src/storage/kv/store.rs
arriqaaq: Totally agree, thank you for catching this

src/storage/kv/store.rs
arriqaaq: You are absolutely correct.

Regarding the commit being async, I had a discussion with Rushmore regarding not having the current thread blocking while waiting for disk I/O. Hence, in the commit phase, the entries are pushed into a commit pipeline, as you mentioned, and once the task is done, we are notified through a channel, and hence marked as completed.

src/storage/kv/store.rs
arriqaaq: There is one issue though, in the [mutex PR](https://github.com/surrealdb/surrealkv/pull/12) I am using async locks, because of which this this has to become async as taking a lock is also async. I am not really sure if I need the async locks though, but wdyt regarding that PR? If not required, would be easy to make this sync

src/storage/kv/store.rs
arriqaaq: Same confusion as with the [async mutex](https://github.com/surrealdb/surrealkv/pull/12) PR

src/storage/kv/store.rs
emmanuel-keller: Indeed, may be even `error!`?

src/storage/kv/store.rs
```diff
@@ -97,15 +173,31 @@ pub struct Core {
...
-    pub(crate) manifest: Aol,
+    pub(crate) manifest: RwLock<Aol>,
     /// Transaction ID Oracle for store.
     pub(crate) oracle: Arc<Oracle>,
     /// Value cache for store.
-    /// The assumption for this cache is that it could be useful for
+    /// The assumption for this cache is that it should be useful for
     /// storing offsets that are frequently accessed (especially in
     /// the case of range scans)
     pub(crate) value_cache: Arc<RwLock<LruCache<u64, Bytes>>>,
```
emmanuel-keller, line 183: In Surreal core, we recently switched from LruCache to QuickCache. Async + concurrent

https://crates.io/crates/quick_cache

emmanuel-keller (approved):

src/storage/kv/store.rs
arriqaaq: Thanks for the suggestion @emmanuel-keller !

arriqaaq: I am using the LRU cache and planning to replace it with the s3-fifo cache for better hit ratios. Will try to benchmark this wrt QuickCache

src/storage/kv/store.rs
arriqaaq: I am not going to include the async mutex PR in this release of async core. Will mark this as sync

.gitignore
```diff
@@ -42,6 +42,7 @@ Temporary Items
 /docker/
 /.direnv/
 /testdata/
+/example/
```
rushmorem, line 45: I'm not sure what kind of example you have locally but have you considered adding it to the `examples` folder and committing it?

Cargo.toml
```diff
@@ -13,23 +13,21 @@ crossbeam-channel = "0.5.8"
...
 lru = "0.12.0"
-
-# below is for the async runtime
-async-task = "4.4.0"
-futures-lite = "1.12.0"
-once_cell = "1.17.1"
-flume = "0.11.0"
+async-channel = "2.1.1"
+futures = "0.3.30"
+async-std = "1.12.0"
```
rushmorem, line 18: Why are we using `async-std` instead of Tokio? We use Tokio in `surrealdb`. This way we will end up with two runtimes, instead of one.

I like `async-std`'s APIs but it's not as well maintained as Tokio. Its latest release is from over a year ago. Tokio is far more popular and a much safer bet as far as maintenance goes.

.gitignore
arriqaaq: Good point, just a left over, removing it

src/storage/kv/store.rs
phughk: I didnt realise we were choosing to not have a WAL. The WAL gives us a lot of benefits, even if it doesnt encompass a full system (i.e. just a shard).

- system backups can be incremental instead of always being full snapshot
- read replicas can lag behind leaders at any rate
- nodes can recover from a baseline snapshot and replay the WAL - this makes rejoining clusters easier
- streaming functionality heavily relies on WAL. Without it, there is no easy way to determine event ordering.

arriqaaq: Actually the transaction log is the ordered immutable log which can be used for most of the above. What I was talking about in terms of WAL is to add durability between precommit to commit. 

I agree, this will help in the replication/failover recovery, I am planning to add it post the async stuff

.gitignore
arriqaaq: Was just thinking of adding examples of how to use it, but will keep it for a separate PR

Cargo.toml
arriqaaq: Removed this dependency and replaced with tokio

src/storage/kv/store.rs
mumoshu: 👍 Will follow #12!

Cargo.toml
arriqaaq: Removed dependency on async_std, thanks for the explanation @rushmorem :)

benches/store_bench.rs
maxwellflitton: `Tokio` will park it's threads if it's not being used. Therefore, you can have two runtimes. If one runtime is no longer being utilised, it will park all it's threads.

src/storage/kv/store.rs
arriqaaq: Also added QuickCache here @emmanuel-keller !

rushmorem (approved):

PR #11, opened 2024-02-03
Remove unneeded usage of MaybeUninit

Ax9D:
The children nodes can be stored as `Option<Arc<N>>` instead of `MaybeUninit<Option<Arc<N>>`.
This way we get rid of unnecessary usage of unsafe when trying to access nodes.

I believe `MaybeUninit` was initially used because it wasn't straightforward to initialize a fixed length array of children. It is now being initialized like this:

```rust
let children: [Option<Arc<N>>; WIDTH] = std::array::from_fn(|_| None);
```

### DIFF
--- a/src/storage/index/node.rs
+++ b/src/storage/index/node.rs
@@ -1,4 +1,3 @@
-use std::mem::MaybeUninit;
 use std::sync::Arc;
 
 use crate::storage::index::{BitArray, KeyTrait};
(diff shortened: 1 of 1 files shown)


arriqaaq (approved):
LGTM!

arriqaaq:
Thank you @Ax9D for this fix, appreciate it. If you are interested, could you also open a PR for this in the [VART](https://github.com/surrealdb/vart) library? The VART is the index code which will actually be used.

Thank you so much, @Ax9D, for taking the time to address this issue. Your effort is truly appreciated. If you're open to it, would you consider submitting a pull request for this fix in the [VART](https://github.com/surrealdb/vart) library? VART is the core code that will be actively utilized (index is just a copy of that).

Ax9D:
Glad I could help!

I've made a PR in VART as well.

PR #12, opened 2024-02-05
feat: switched to async mutex

arriqaaq:
This PR adapts usage of async locks by switching from parking_lot mutex to tokio::sync::async mutex.


src/storage/kv/oracle.rs
```diff
@@ -400,8 +408,8 @@ impl Mark {
         })
     }
 
-    fn take(&self) -> Sender<()> {
-        self.ch.lock().take().unwrap()
+    async fn take(&self) -> Sender<()> {
```
phughk, line 411: We should allow synchrnonous alternatives alongside the async default ones. `try_take` and `take_blocking` for example. But that can be in a separate PR obviously, or even further down the line. I just know they will be useful to avoid sync/async interleaving.

phughk (approved):
Awesome lgtm! Thanks

emmanuel-keller:
General thought: How do the sync and the async benchmarks compare? Is the level of performance identical?

arriqaaq:
> General thought: Now that we have both a sync and an async version of the same benchmark, how do they compare? Is the level of performance identical?

@emmanuel-keller Here is the difference between existing `main` branch vs `async` (without using async locks)

```
(async)
insert key/value lengths 10/10
                        time:   [15.230 µs 15.474 µs 15.729 µs]
sequential inserts      time:   [180.34 ns 180.76 ns 181.21 ns]
sequential gets         time:   [2.8562 ns 2.8595 ns 2.8629 ns]

(sync)
insert key/value lengths 10/10
                        time:   [5.6125 µs 5.8483 µs 6.0650 µs]
sequential inserts      time:   [49.604 µs 50.466 µs 51.194 µs]
sequential gets         time:   [504.90 ps 518.60 ps 536.32 ps]
```

Don't see any improvements as such, except for the sequential insert. Need to check if it is an implementation issue or not.

arriqaaq:
> > General thought: Now that we have both a sync and an async version of the same benchmark, how do they compare? Is the level of performance identical?
> 
> @emmanuel-keller Here is the difference between existing `main` branch vs `async` (without using async locks)
> 
> ```
> (async)
> insert key/value lengths 10/10
>                         time:   [15.230 µs 15.474 µs 15.729 µs]
> sequential inserts      time:   [180.34 ns 180.76 ns 181.21 ns]
> sequential gets         time:   [2.8562 ns 2.8595 ns 2.8629 ns]
> 
> (sync)
> insert key/value lengths 10/10
>                         time:   [5.6125 µs 5.8483 µs 6.0650 µs]
> sequential inserts      time:   [49.604 µs 50.466 µs 51.194 µs]
> sequential gets         time:   [504.90 ps 518.60 ps 536.32 ps]
> ```
> 
> Don't see any massive improvements as such, except for the sequential insert. Need to check if it is an implementation issue or not.

Concurrent benchmark

```
(sync)
inserts/100000 inserts (1 threads)
                        time:   [863.39 ms 885.87 ms 911.93 ms]
                        thrpt:  [109.66 Kelem/s 112.88 Kelem/s 115.82 Kelem/s]
inserts/100000 inserts (2 threads)
                        time:   [1.1645 s 1.1830 s 1.2031 s]
                        thrpt:  [83.119 Kelem/s 84.530 Kelem/s 85.874 Kelem/s]
inserts/100000 inserts (4 threads)
                        time:   [1.4087 s 1.4849 s 1.5828 s]
                        thrpt:  [63.180 Kelem/s 67.343 Kelem/s 70.987 Kelem/s]

(async)
inserts/100000 inserts (1 threads)
                        time:   [1.9925 s 2.0181 s 2.0576 s]
                        thrpt:  [48.601 Kelem/s 49.552 Kelem/s 50.188 Kelem/s]
inserts/100000 inserts (2 threads)
                        time:   [1.6085 s 1.6430 s 1.6797 s]
                        thrpt:  [59.533 Kelem/s 60.863 Kelem/s 62.170 Kelem/s]
inserts/100000 inserts (4 threads)
                        time:   [1.8308 s 1.9657 s 2.1281 s]
                        thrpt:  [46.990 Kelem/s 50.873 Kelem/s 54.621 Kelem/s]

```

src/storage/kv/oracle.rs
arriqaaq: I agree with you, I am just keeping this as an optional PR, will work on what you said in this

src/storage/kv/entry.rs
```diff
@@ -276,8 +277,9 @@ impl TxEntry {
     }
 }
 
+#[async_trait]
```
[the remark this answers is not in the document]
arriqaaq: Ah I wasn't aware. Thank you for the info

PR #17, opened 2024-02-09
fix: do not delete key from snapshot until committed

arriqaaq:
## Description

This PR fixes the deletion bug in the store due to deletion of keys from snapshot before committing it and adds tests to verify the behavior of transactions when keys are deleted and then scanned. 

## Test Strategy
The test first creates a key-value pair in a read-write transaction, then deletes the key in a read-only transaction and scans the range where the key was located. The test asserts that the scan does not return the deleted key.

### DIFF
--- a/src/storage/kv/transaction.rs
+++ b/src/storage/kv/transaction.rs
@@ -221,13 +221,8 @@ impl Transaction {
             // Convert the value to Bytes.
             let index_value = ValueRef::encode_mem(&e.value, e.metadata.as_ref());
 
-            // If the entry is marked as deleted, delete it from the snapshot.
-            // Otherwise, set the key-value pair in the snapshot.
-            if e.is_deleted() {
-                self.snapshot.write().delete(&e.key[..].into())?;
-            } else {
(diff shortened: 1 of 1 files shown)


phughk (approved):
Approved, but surprised by the fix. Discussing in slack kv channel

src/storage/kv/transaction.rs
```diff
@@ -1493,4 +1488,107 @@ mod tests {
...
+            let mut txn1 = store.begin().unwrap();
+            txn1.set(&key1, &value1).unwrap();
+            txn1.commit().await.unwrap();
+        }
+
+        {
+            // Start a read-only transaction (txn)
+            let mut txn = store.begin().unwrap();
+            txn.get(&key1).unwrap();
+            txn.delete(&key1).unwrap();
```
phughk, line 1511: Shouldn't this operation fail under R-only?

arriqaaq: This is a transaction that commits, do you mean this has to be Read only?

rushmorem (approved):

PR #18, opened 2024-02-09
Import vart and bump version

arriqaaq:
## Description

This PR addresses the following isssues:

1. Fixes an issue in reading deleted keys within a transaction. Previously, when a key was deleted within a transaction and you try to read it (within the transaction itself, ryow), the system would return Some([]) instead of None. This behavior was misleading, as it suggested that the key was present but had an empty value, rather than being deleted. With this fix, the system now correctly returns None for deleted keys when doing ryow (read your own writes), accurately reflecting their absence in the snapshot. 
2. Bumps up the crate version to release the fix
3. Imports vart and replaces index
4. Fixes clippy warnings


mumoshu:
@arriqaaq Can you maybe point me which commit in `vart` corresponds to the first fix about reading deleted keys?!

From reading the changeset it looks to me that the fix is coming solely from migrating from the internal art package to the new `vart`.

src/storage/kv/transaction.rs
```diff
@@ -157,7 +158,11 @@ impl Transaction {
                 // Check if the key is in the write set by checking in the write_order_map map.
                 if let Some(order) = self.write_order_map.get(&hashed_key) {
                     if let Some(entry) = self.write_set.get(*order as usize) {
-                        return Ok(Some(entry.1.value.clone().to_vec()));
+                        return Ok(if entry.1.is_deleted() {
```
arriqaaq, line 161: @mumoshu This is actually the place. So when an entry is deleted, I store empty value against it in the snapshot, but when I was reading it, it used to return Some([]) when it had to return None.

tobiemh (approved):

PR #24, opened 2024-02-22
fix: reset cursor to eof on append

arriqaaq:
## Description

When the same file handler is being used, once a read happens, the cursor needs to be set to the end of file before appending again to avoid override.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "surrealkv"
 publish = true
-version = "0.1.2"
+version = "0.1.3"
 edition = "2021"
 license = "Apache-2.0"
 readme = "README.md"
--- a/src/storage/log/aof/log.rs
+++ b/src/storage/log/aof/log.rs
@@ -419,7 +419,7 @@ mod tests {
     }
 
     #[test]
-    fn append_read_append() {
+    fn append_read_append_read() {
         // Create a temporary directory
         let temp_dir = create_temp_directory();
 
--- a/src/storage/log/aof/log.rs
+++ b/src/storage/log/aof/log.rs
@@ -378,4 +378,100 @@ mod tests {
         // Test closing segment
         assert!(a.close().is_ok());
     }
+
+    #[test]
+    fn append_and_read_two_blocks() {
+        // Create a temporary directory
+        let temp_dir = create_temp_directory();
+
+        // Create aol options and open a aol file
+        let opts = Options::default();
+        let mut a = Aol::open(temp_dir.path(), &opts).expect("should create aol");
+
+        // Create two slices of bytes of different sizes
+        let data1 = vec![1; 31 * 1024];
+        let data2 = vec![2; 2 * 1024];
+
+        // Append the first data slice to the aol
+        let r1 = a.append(&data1);
+        assert!(r1.is_ok());
+        assert_eq!(31 * 1024, r1.unwrap().1);
+
+        // Append the second data slice to the aol
+        let r2 = a.append(&data2);
+        assert!(r2.is_ok());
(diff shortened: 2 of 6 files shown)


emmanuel-keller (approved):
LGTM!

tobiemh (approved):

src/storage/log/mod.rs
```diff
@@ -712,6 +712,12 @@ fn merge_slices(dest: &mut [u8], src: &[u8]) -> usize {
     min_len
 }
 
+fn copy_into_dest_from_src(dest: &mut [u8], dest_len: usize, src: &[u8], src_len: usize) -> usize {
```
mumoshu, line 715: Nit but either naming this so that it is clear that "it tries to copy dest.len bytes. but no more than src_len, from src"(hope it makes sense 😄), or add a comment to this function to explain that would be easier to read!

src/storage/log/aof/log.rs
```diff
@@ -378,4 +378,100 @@ mod tests {
...
+        let n2 = a.read_at(&mut read_data2, 31 * 1024).expect("should read");
+        assert_eq!(2 * 1024, n2);
+        assert_eq!(data2, read_data2);
+
+        // Test closing segment
+        assert!(a.close().is_ok());
+    }
+
+    #[test]
+    fn append_read_append() {
```
mumoshu, line 422: Maybe name this `append_read_append_read` or `append_append`(assuming each read is meant to verify the previous append, read itself is not a test subject) if that makes sense to you too?

src/storage/log/aof/log.rs
```diff
@@ -378,4 +378,100 @@ mod tests {
         // Test closing segment
         assert!(a.close().is_ok());
     }
+
+    #[test]
+    fn append_and_read_two_blocks() {
```
mumoshu, line 383: My quick thought was that `append_read_append` is enough for verifying this issue.
Would you mind clarifying why we need both tests? Is reading two blocks an important distinction?

src/storage/log/aof/log.rs
arriqaaq: Hey @mumoshu 

So there is one file handler, and this test checks if post append and read, the cursor is not shifted back hence wanted to explicitly mention the read part. Will change it to `append_read_append_read `

src/storage/log/aof/log.rs
arriqaaq: So I am writing in block size of 32kb to disk. There was an issue where if one full block is written, and other one is partially written to, read shouldn't fail spanning two blocks (one on disk and other in memory)

src/storage/log/mod.rs
arriqaaq: Ack! Will do so

arriqaaq: Added comment, lemme know if this is ok

src/storage/log/aof/log.rs
arriqaaq: [This](https://github.com/surrealdb/surrealkv/blob/0725edbedcf5ac55918abd93c26a05d53f309181/src/storage/log/mod.rs#L1162) should provide more context

phughk:
Got merged mid-review 😂  Would have approved anyway though. Thanks!

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "surrealkv"
 publish = true
-version = "0.1.2"
+version = "0.1.3"
```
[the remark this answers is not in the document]
arriqaaq: Noted @phughk, I will do that, thank you for highlighting

arriqaaq:
Thank you @phughk, will keep note of the PR for version bump

PR #30, opened 2024-03-13
fix: close_reopen test should read inserted data and append in next run

arriqaaq:
## Description

This test case ensures that once data is added to skv, upon each reopening, it can successfully read the data and then append to it accordingly without any errors.

### DIFF
--- a/src/storage/kv/store.rs
+++ b/src/storage/kv/store.rs
@@ -608,16 +608,17 @@ mod tests {
         let mut opts = Options::new();
         opts.dir = temp_dir.path().to_path_buf();
 
-        let num_ops = 1000;
+        let num_ops = 10;
 
         for i in 1..=10 {
             // (Re)open the store
             let store = Store::new(opts.clone()).expect("should create store");
 
             // Append num_ops items to the store
-            for j in (num_ops * (i - 1))..(num_ops * i) {
-                let key = format!("key{}", j);
-                let value = format!("value{}", j);
+            for j in 0..num_ops {
+                let key = format!("key{}", (i - 1) * num_ops + j);
+                let value = (i - 1) * num_ops + j;
+                let value = format!("value{}", value);
                 let mut txn = store.begin().unwrap();
                 txn.set(key.as_bytes(), value.as_bytes()).unwrap();
                 txn.commit().await.unwrap();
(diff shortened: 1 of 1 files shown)


src/storage/kv/store.rs
```diff
@@ -601,36 +601,32 @@ mod tests {
...
+        for i in 1..=10 {
+            // (Re)open the store
             let store = Store::new(opts.clone()).expect("should create store");
 
-            // Insert 1000 items into the store
-            for i in 0..1000 {
-                let key = format!("key{}", i);
-                let value = format!("value{}", i);
+            // Append num_ops items to the store
+            for j in (num_ops * (i - 1))..(num_ops * i) {
```
sgirones, line 618: for visibility purposes, I would move this math to a temp var, because it's not actually relevant for the loop
```
for j in 0..num_ops {
  let val = (i - 1) * num_ops + j
  let key = format!("key{}", val);
  let value = format!("value{}", val);
  ...
}
```

src/storage/kv/store.rs
```diff
@@ -601,36 +601,32 @@ mod tests {
...
-            store.close().await.unwrap();
-
-            // Reopen the store
-            let store = Store::new(opts.clone()).expect("should create store");
-
             // Test that the items are still in the store
-            for i in 0..1000 {
-                let key = format!("key{}", i);
-                let value = format!("value{}", i);
+            for j in 0..(num_ops * i) {
```
sgirones, line 627: you do a full check since the first key on every iteration? is this intentional?

src/storage/kv/store.rs
arriqaaq: ack, done

src/storage/kv/store.rs
```diff
@@ -601,36 +601,33 @@ mod tests {
...
 
-            // Insert 1000 items into the store
-            for i in 0..1000 {
-                let key = format!("key{}", i);
-                let value = format!("value{}", i);
+            // Append num_ops items to the store
+            for j in 0..num_ops {
+                let key = format!("key{}", (i - 1) * num_ops + j);
+                let value = (i - 1) * num_ops + j;
+                let value = format!("value{}", value);
```
sgirones, line 621:
```suggestion
                let val = (i - 1) * num_ops + j;
                let key = format!("key{}", val);
                let value = format!("value{}", val);
```

sgirones (approved):

PR #31, opened 2024-03-13
[store] Close the store asynchronously when it's dropped

sgirones:
The `store.close()` method is async so it can't be directly called from `fn drop()`.

This PR introduces a Store wrapper that handles the drop asynchronously.

There's the caveat that if the Store is dropped and then we immediately try to open it again, the original store may not be closed yet.

In order to overcome that, the user can explicitly close the store instead of relying on the drop.

The drop is mainly intended to handle the common scenario where the Store is dropped when the program exits.

### DIFF
--- a/src/storage/kv/store.rs
+++ b/src/storage/kv/store.rs
@@ -93,8 +93,6 @@ impl StoreInner {
 #[derive(Default)]
 pub struct Store {
     pub(crate) inner: Option<StoreInner>,
-    // Flag to indicate if the store has been dropped so we don't enter a loop
-    pub(self) _dropped: bool,
 }
 
 impl Store {
@@ -102,7 +100,6 @@ impl Store {
     pub fn new(opts: Options) -> Result<Self> {
         Ok(Self {
             inner: Some(StoreInner::new(opts)?),
-            _dropped: false,
         })
     }
 
@@ -158,15 +155,10 @@ impl Store {
 
 impl Drop for Store {
     fn drop(&mut self) {
-        if !self._dropped {
-            // Replace the current self with a default, so we can send the original self to the async task
-            let mut this = Store::default();
-            std::mem::swap(&mut this, self);
-            this._dropped = true;
-
+        if let Some(inner) = self.inner.take() {
             // Close the store asynchronously
             tokio::spawn(async move {
-                if let Err(err) = this.close().await {
+                if let Err(err) = inner.close().await {
                     // TODO: use log/tracing instead of eprintln
                     eprintln!("Error occurred while closing the kv store: {}", err);
                 }
@@ -545,7 +537,6 @@ impl Core {
 
 #[cfg(test)]
 mod tests {
-    use std::borrow::Borrow;
     use std::sync::Arc;
 
     use crate::storage::kv::option::Options;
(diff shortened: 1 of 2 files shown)


src/storage/kv/store.rs
```diff
@@ -97,30 +146,32 @@ impl Store {
...
-                    e
-                ))
-            })?;
-        }
+impl Drop for Store {
+    fn drop(&mut self) {
+        if !self._dropped {
+            // Replace the current self with a default, so we can send the original self to the async task
+            let mut this = Store::default();
+            std::mem::swap(&mut this, self);
```
phughk, line 164: That's really cool, I didn't realise this was possible

src/storage/kv/store.rs
```diff
@@ -97,30 +146,32 @@ impl Store {
...
+            this._dropped = true;
 
-        self.core.close()?;
-        Ok(())
+            // Close the store asynchronously
+            tokio::spawn(async move {
+                if let Err(err) = this.close().await {
```
phughk, line 169: Im wondering if it does actually make sense to have this. I know the bug is that there is termination before a proper shutdown. But surely we should be building to tolerate such failures, instead of building to hope that this async drop is successful?

phughk (approved):
Approving the code, but wondering if it makes sense to have this

sgirones:
Absolutely. This one in isolation gives no guarantees, we need something else.

Having this though allows for implicit graceful shutdown.

src/storage/kv/store.rs
```diff
@@ -97,30 +146,32 @@ impl Store {
...
-                    "Error occurred while closing the kv store. JoinError: {}",
-                    e
-                ))
-            })?;
-        }
+impl Drop for Store {
+    fn drop(&mut self) {
+        if !self._dropped {
+            // Replace the current self with a default, so we can send the original self to the async task
+            let mut this = Store::default();
```
arriqaaq, line 163: WDYT about just using take on the inner field because it is an option? That way, we wouldn't need the _dropped field right?

```
let this = self.inner.take().unwrap();
// Close the store asynchronously
tokio::spawn(async move {
    if let Err(err) = this.close().await {
        // TODO: use log/tracing instead of eprintln
        eprintln!("Error occurred while closing the kv store: {}", err);
    }
});

```

src/storage/kv/store.rs
```diff
@@ -97,30 +146,32 @@ impl Store {
...
+            std::mem::swap(&mut this, self);
+            this._dropped = true;
 
-        self.core.close()?;
-        Ok(())
+            // Close the store asynchronously
+            tokio::spawn(async move {
+                if let Err(err) = this.close().await {
+                    // TODO: use log/tracing instead of eprintln
+                    eprintln!("Error occurred while closing the kv store: {}", err);
```
arriqaaq, line 171: WDYT about panic here? Because this is such a critical part, just silently printing as an error inside sdb could go unnoticed no?

src/storage/kv/store.rs
sgirones: ah, good one!

src/storage/kv/store.rs
sgirones: Yeah, was on the fence here. But a Panic will abort the process and potentially affect other tasks that had a chance to finish correctly.

As a general rule of thumb, we should never panic in a release binary, but maybe is a good idea to panic for debug binaries

arriqaaq (approved):
LGTM!

PR #33, opened 2024-03-21
Feat/add durability option at transaction level

arriqaaq:
## Description

This PR adds a durability option per transaction. This option enables a user to choose to commit a transaction to disk immediately or batch it in BLOCK_SIZE


src/storage/log/aof/log.rs
```diff
@@ -274,6 +290,7 @@ impl Aol {
     }
 
     pub fn close(&mut self) -> Result<()> {
+        self.check_if_fsync_failed()?;
```
mumoshu, line 293: Does this mean we cannot close the Aol even though the comment on `check_if_fsync_failed` says you must close the store and restart?
[It seems like the store needs to close the aol in order to close itself.](https://github.com/surrealdb/surrealkv/blob/aca3678480fcd4329b009da584cee43da3a21a03/src/storage/kv/store.rs#L445-L446)

Can we perhaps say you cannot gracefully close the store(along with aol) when there was a fsync failure, and instead you need to forcefully restart the store somehow, and we need to implement a way to restart the store forcefully?

src/storage/log/mod.rs
```diff
@@ -17,7 +17,7 @@ use hashbrown::HashMap;
 /// The `BLOCK_SIZE` constant represents the size of a block used for buffering disk writes in the
 /// write-ahead log. It determines the maximum amount of data that can be held in memory before
 /// flushing to disk. Larger block sizes can improve write performance but might use more memory.
-pub(crate) const BLOCK_SIZE: usize = 32768;
+pub(crate) const BLOCK_SIZE: usize = 4 * 1024;
```
mumoshu, line 20: Was there any specific reason we wanted to decrease the max aol segment size(correct?) from 32KB down to 4KB here?

I think this is only for `Durability::Eventual` and does not affect `Immediate`, and you usually want Eventual when you need more throughput in exchange of durability, which justifies larger segment size?

src/storage/log/aof/log.rs
```diff
@@ -474,4 +511,57 @@ mod tests {
...
+        assert_eq!(1024, a.offset().unwrap());
+
+        // Simulate fsync failure
+        a.set_fsync_failed(true);
+
+        // Writes should fail aftet fsync failure
+        let r = a.append(&small_record);
+        assert!(r.is_err());
+
+        // Reads should fai after fsync failure
```
mumoshu, line 562:
```suggestion
        // Reads should fail after fsync failure
```

mumoshu (approved):
A potential issue in the aol close implementation, but LGTM overall!
Thank you very much for your awesome work @arriqaaq!

src/storage/log/aof/log.rs
arriqaaq: Thank you for having a detailed look @mumoshu !

You raise a valid point, I think for close this check is not needed as you have rightly mentioned because we need to fsync to update the metadata of the directory during close. I am trying to repair on restart as you've reviewed in this [draft PR](https://github.com/surrealdb/surrealkv/pull/35). I'll try to remove this check.

src/storage/log/mod.rs
arriqaaq: So the reason I did not want to keep 32kb is because as per my observation, a sdb transaction was occupying around ~1000 bytes. I can't tell this with certainty for sure because it depends on the type of key/values being used. But the issue was that there could be a loss of too many transactions in case of an abrupt restart if the size is small. Hence I decided to keep it small, and probably make this configurable with time to improve performance.

If Durability is eventual, the transactions are just buffered as per the BLOCK_SIZE as described [here](https://github.com/surrealdb/surrealkv/blob/aca3678480fcd4329b009da584cee43da3a21a03/src/storage/log/mod.rs#L1022) and then flushed to disk. The segment size is to avoid keeping a single file for commit log, this way it is easier to do chunk upload to s3 and also for recovery, so that was the motive behind writing in chunks. I've increased this to 512mb in subsequent PRs.

Please let me know if that answers your question

arriqaaq: Reverted back to 32kb

PR #35, opened 2024-03-21
Feat/add repair for commit log

arriqaaq:
## Description

This PR adds a repair method to recover from the corruption of the commit log due to checksum errors or partial writes failure.


src/storage/kv/repair.rs
```diff
@@ -0,0 +1,540 @@
...
+    // Open a new segment as the active segment
+    let mut new_segment: Segment<0> = Segment::open(&aol.dir, corrupted_segment_id, &aol.opts)?;
+
+    // Create a segment reader for the repaired segment
+    let segments: Vec<SegmentRef> = vec![SegmentRef {
+        file_path: repaired_segment_path.clone(),
+        file_header_offset: corrupted_segment_file_header_offset,
+        id: corrupted_segment_id,
+    }];
+    let segment_reader = MultiSegmentReader::new(segments)?;
```
mumoshu, line 72: It's not strictly needed feature-wise, but I'm a bit concerned that this function is tightly coupled to the `log` mod's internals.

Can we probably extract L30-L71 into a function within the log mod and call that from here? That way, there will be more separation of concerns.

All in all, I might imagine something like the below here:

```
let segments = SegmentRef::repair_and_read_segments_from_directory(aol, corrupted_segment_id)?;
let segment_reader = MultiSegmentReader::new(segments)?;
```

arriqaaq: Thanks for the suggestion @mumoshu ! I'll look into how this can be resolved

src/storage/kv/repair.rs
```diff
@@ -0,0 +1,839 @@
...
+        // If the segment ID matches the corrupted segment ID, repair the segment
+        if s.id == corrupted_segment_id {
+            file_path = Some(s.file_path.clone());
+            file_header_offset = s.file_header_offset;
+            break;
+        }
+    }
+
+    // Explicitly drop the segments here to ensure no file descriptors are kept
+    drop(segs);
```
emmanuel-keller, line 82: NIT: Another way of ensuring that `segs` is dropped is using a scope.
When `segs` is out of scope it will be dropped.

src/storage/log/aof/log.rs
```diff
@@ -146,11 +146,10 @@ impl Aol {
 
         // Check if the record is larger than the maximum file size
         if rec.len() > self.opts.max_file_size as usize {
-            eprintln!("rec len {} {}", rec.len(), self.opts.max_file_size);
             return Err(Error::RecordTooLarge);
         }
 
-        let _lock = self.mutex.write();
+        let _lock = self.mutex.lock();
```
emmanuel-keller, line 152: I would suggest to rename _lock, lock and use drop later when you want to be sure the lock id dropped.
Otherwise I suspect that a compiler optimisation may release the lock early.

mumoshu: I wasn't sure myself at first, but after some research, I realized this is a valid usage!
A variable definition automatically adds a drop at the scope end regardless of whether it's prefixed with _ or not.
The compiler won't remove it as an optimization because, if it did, it would change the program's observable behavior (lock taken/released).

mumoshu: Regardless, I second the idea of using e.g. RwLock as Emmanuel has suggested, because it makes us not to accidentally miss locking when mutating active_segment.

src/storage/log/aof/log.rs
```diff
@@ -298,19 +297,19 @@ impl Aol {
...
         self.active_segment.close()?;
         Ok(())
     }
 
     // Returns the current offset within the segment.
     pub fn offset(&self) -> Result<u64> {
-        let _lock = self.mutex.read();
+        let _lock = self.mutex.lock();
         Ok((self.active_segment_id * self.opts.max_file_size) + self.active_segment.offset())
```
mumoshu, line 308:
```suggestion
        Ok(self.calculate_offset() + self.active_segment.offset())
```

Probably calculate_offset is handy here too?

src/storage/kv/repair.rs
arriqaaq: Thanks @emmanuel-keller for the tip! Done

src/storage/log/aof/log.rs
arriqaaq: Thanks @emmanuel-keller and @mumoshu for the details! I will tackle this in a separate PR to make lock exclusive for the active segment

src/storage/log/aof/log.rs
arriqaaq: Thanks @mumoshu !

phughk (approved):
Lgtm thanks!

arriqaaq:
Thank you for your valuable comments and review @emmanuel-keller @mumoshu @phughk !

PR #38, opened 2024-04-16
feat: add option to store db in memory

arriqaaq:
## Description

This PR adds the ability for skv to act as a pure in-memory store. This can be used as alternative to other in-memory stores whilst testing inside sdb

### DIFF
--- a/src/storage/kv/store.rs
+++ b/src/storage/kv/store.rs
@@ -322,20 +322,6 @@ impl Core {
             }
         }
 
-        // // Determine options for the manifest file and open or create it.
-        // let mut manifest = Self::initialize_manifest(&opts)?;
-
-        // // Load options from the manifest file.
-        // let opts = Core::load_options(&opts, &mut manifest)?;
-
-        // // Determine options for the commit log file and open or create it.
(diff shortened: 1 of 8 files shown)


src/storage/kv/store.rs
```diff
@@ -304,20 +303,39 @@ impl Core {
         // Initialize a new Indexer with the provided options.
         let mut indexer = Self::initialize_indexer();
 
-        // Determine options for the manifest file and open or create it.
-        let mut manifest = Self::initialize_manifest(&opts)?;
+        let mut manifest = None;
+        let mut clog = None;
+
+        if opts.should_persist_data() {
```
mumoshu, line 309: Does this mean we omit persistence at all, rather than storing clog in memory?

// I was wondering if we want to have a way to read clog using the client, so that it can be leveraged for e.g. CF (and hence LQ), or even potentially logical incremental backups/snapshots etc.

Not thought throughout yet, but if there's any possibility we can maintain clog in-memory rather than omitting clog at all, that would be great!

mumoshu (approved):
LGTM!

PR #42, opened 2024-05-02
Fix/watermark fetch

arriqaaq:
## Description

Fixes a bug where the watermark could panic if a waiter is not present


phughk (approved):
Lgtm thanks!

PR #47, opened 2024-06-25
Feat/add clear method

arriqaaq:
## Description

This PR does the following: 

- Adds a `clear` method to the transaction. The difference between `clear` and `delete` is that `clear` does a soft delete but `delete` hard deletes the key with all its versions.
- Fixes compaction to use versioned iterator from vart to record all versions during compaction


emmanuel-keller (approved):
Welcome clear!

PR #50, opened 2024-07-02
chore: add methods to get keys at specific timestamps

arriqaaq:
## Description

This PR does the following:

1) Updates vart
2) Adds the following versioned APIs to the snapshot and tree
    - `get_at_ts` (to get a key at a particular timestamp)
    - `get_all_versions` (to get all the versions of a key)
    - `scan_at_ts` (scans all keys at a timestamp)
    - `keys_at_ts` (query all keys at a timestamp)
3) Add `get_value_by_query` to both tree and snapshot to query the value by the following query types
    - `LatestByVersion`
    - `LatestByTs`
    - `LastLessThanTs`
    - `LastLessOrEqualTs`
    - `FirstGreaterThanTs`
    - `FirstGreaterOrEqualTs`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ bytes = "1.5.0"
 tokio = { version = "1.36", features = ["rt", "sync"] }
 sha2 = "0.10.8"
 quick_cache = "0.5.1"
-vart = "0.3.1"
+vart = {git = "https://github.com/surrealdb/vart", branch = "chore/add-versioned-apis"}
 revision = "0.7.1"
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ bytes = "1.5.0"
 tokio = { version = "1.36", features = ["rt", "sync"] }
 sha2 = "0.10.8"
 quick_cache = "0.5.1"
-vart = {git = "https://github.com/surrealdb/vart", branch = "chore/add-versioned-apis"}
+vart = {git = "https://github.com/surrealdb/vart", branch = "main"}
 revision = "0.7.1"
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ bytes = "1.5.0"
 tokio = { version = "1.36", features = ["rt", "sync"] }
 sha2 = "0.10.8"
 quick_cache = "0.5.1"
-vart = {git = "https://github.com/surrealdb/vart", branch = "main"}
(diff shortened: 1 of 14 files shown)


src/storage/kv/store.rs
```diff
@@ -172,13 +175,50 @@ impl Store {
...
+    /// Returns the value associated with the key.
+    pub fn get(&self, key: &[u8]) -> Result<Vec<u8>> {
+        let core = self.inner.as_ref().unwrap().core.clone();
+        let snapshot = Snapshot::take(core, now())?;
+        let val_ref = snapshot.get(&key[..].into())?;
+        val_ref.resolve()
+    }
+
+    /// Returns the value associated with the key at the given timestamp.
+    pub fn get_at_ts(&self, key: &[u8], ts: u64) -> Result<Vec<u8>> {
```
mumoshu, line 196: Same as https://github.com/surrealdb/surrealkv/pull/50/files#r1662192392

src/storage/kv/store.rs
```diff
@@ -172,13 +175,50 @@ impl Store {
...
     pub async fn compact(&self) -> Result<()> {
         if let Some(inner) = self.inner.as_ref() {
             inner.compact().await?;
         }
 
         Ok(())
     }
+
+    /// Returns the value associated with the key.
+    pub fn get(&self, key: &[u8]) -> Result<Vec<u8>> {
```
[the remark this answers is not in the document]
arriqaaq: Totally agree with you Yusuke, I wanted to add these for non-surrealdb users so that if someone just wants to query the store without having to worry about transaction, they could do so. Also I was confused about how these version APIs would play into the transactional integrity, but given that we are now adding it to read-only transactions, I will remove this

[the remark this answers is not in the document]
arriqaaq: Removed!

src/storage/kv/store.rs
arriqaaq: Removed!

Cargo.toml
```diff
@@ -29,7 +29,7 @@ bytes = "1.5.0"
 tokio = { version = "1.36", features = ["rt", "sync"] }
 sha2 = "0.10.8"
 quick_cache = "0.5.1"
-vart = "0.3.1"
+vart = {git = "https://github.com/surrealdb/vart", branch = "main"}
```
arriqaaq, line 32: This is replaced with the correct version of vart in [this](https://github.com/surrealdb/surrealkv/pull/51) subsequent PR

gsserge (approved):

PR #60, opened 2024-07-24
Reduce allocations and lookups in the write path

gsserge:
- Make transaction's `write_set` a vector of entries instead of a vector of tuples, since entry already contains key.
- On commit, extract the `write_set` with `mem::take()` instead of cloning.
- Use `HashMap::entry()` instead of a sequence of `get/insert` when checking if the key already exists in `write_order_map`.
- Remove the `Records` struct which was merely an intermediate vector of `Record`s for the encoding of entries.
- Replace `hashbrown` with `ahash`.

### DIFF
--- a/src/storage/kv/transaction.rs
+++ b/src/storage/kv/transaction.rs
@@ -1,9 +1,10 @@
 use std::ops::{Bound, RangeBounds};
 use std::sync::Arc;
 
+use ahash::{HashMap, HashMapExt};
 use bytes::{Bytes, BytesMut};
-use hashbrown::HashMap;
 use parking_lot::{Mutex, RwLock};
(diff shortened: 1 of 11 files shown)


src/storage/kv/transaction.rs
```diff
@@ -2,7 +2,7 @@ use std::ops::{Bound, RangeBounds};
 use std::sync::Arc;
 
 use bytes::{Bytes, BytesMut};
-use hashbrown::HashMap;
+use hashbrown::{hash_map::Entry as HashEntry, HashMap};
```
emmanuel-keller, line 5: Shall we take this opportunity to replace hashbrown as well?

emmanuel-keller (approved):
Nice!

tobiemh (approved):

PR #62, opened 2024-07-25
Remove unused WAL

gsserge:
Historically SurrealKV contained two log implementations, WAL and AOL. Only AOL is currently used, so it makes sense to remove WAL to simplify the codebase. If needed, the WAL implementation is still available in Git history.

@arriqaaq I've removed a few segments tests in `storage/log/mod.rs` because they were using WAL record header, and I'm not completely sure that that was the right call. Please let me know if I should keep them and use AOL segments instead.


arriqaaq (approved):
LGTM! Thanks @gsserge, and as you said correctly, those tests with WAL headers are not required

PR #63, opened 2024-07-26
Remove block buffering from the write path

gsserge:
The `Weak` durability level is not really useful for our use case and can be removed. With it gone, there's no need in buffering writes either, since each record must be fully written into the segment's file descriptor and flushed to the OS buffers. This makes the write path simpler, and also reduces memory usage and copying.

What this PR does:
- Removes `Weak` durability level.
- Removes `Block` struct.
- Removes `flush()` methods because each record is flushed to the OS buffers by default.
- Uses `FileExt::read_at()` where available instead of seeking to and then reading the record.

Tested locally with surrealdb using `cargo make ci-api-integration-surrealkv`.

The block buffering is still used when loading the segments; this will be addressed in a follow-up PR.

### DIFF
--- a/src/storage/log/mod.rs
+++ b/src/storage/log/mod.rs
@@ -10,11 +10,8 @@ use std::io::{self, BufRead, Read, Seek, SeekFrom, Write};
 use std::path::{Path, PathBuf};
 use std::sync::PoisonError;
 
-/// The size of a single block in bytes.
-///
-/// The `BLOCK_SIZE` constant represents the size of a block used for buffering disk writes in the
-/// write-ahead log. It determines the maximum amount of data that can be held in memory before
-/// flushing to disk. Larger block sizes can improve write performance but might use more memory.
+/// The `BLOCK_SIZE` constant represents the size of a block used for buffering disk reads from
+/// the append-only log.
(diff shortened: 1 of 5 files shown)


src/storage/log/mod.rs
```diff
@@ -10,11 +10,8 @@ use std::io::{self, BufRead, Read, Seek, SeekFrom, Write};
...
 use std::sync::PoisonError;
 
-/// The size of a single block in bytes.
-///
-/// The `BLOCK_SIZE` constant represents the size of a block used for buffering disk writes in the
-/// write-ahead log. It determines the maximum amount of data that can be held in memory before
-/// flushing to disk. Larger block sizes can improve write performance but might use more memory.
+/// The `BLOCK_SIZE` constant represents the size of a block used for buffering disk reads from
+/// the append-only log.
 pub(crate) const BLOCK_SIZE: usize = 32 * 1024;
```
arriqaaq, line 15: We won't need this constant now right?

arriqaaq (approved):
LGTM! One minor comment

src/storage/log/mod.rs
gsserge: For this one we still need this. I've got another PR waiting for this to get merged, where I will remove it.

PR #64, opened 2024-07-27
Remove extra read buffering

gsserge:
`MultiSegmentReader` already uses `BufReader` for buffering file reads in memory, therefore there's no need for `Reader` to also buffer reads manually with `Vec<u8>`, and all of the associated code and copying.

Tested locally with surrealdb using `cargo make ci-api-integration-surrealkv`.


tobiemh (approved):

arriqaaq (approved):
LGTM

PR #65, opened 2024-07-31
fix: make checksum based on the entire record fields

arriqaaq:


src/storage/kv/entry.rs
```diff
@@ -134,18 +144,29 @@ impl Record {
...
+        hasher.finalize()
+    }
+
     pub(crate) fn new_from_entry(entry: Entry, tx_id: u64, commit_ts: u64) -> Self {
-        Record {
+        let rec = Record {
             id: tx_id,
             ts: commit_ts,
-            crc32: entry.crc32(),
+            crc32: 0, // Temporarily set to 0, will be updated after initialization
```
gsserge, line 157: Do you it's possible to `impl Hash` for `Entry` instead of `Record` so that there won't be need in initialising `crc32` with `0` first and then assigning it?

gsserge (approved):

src/storage/kv/entry.rs
arriqaaq: That is a really nice suggestion. I was trying to implement it, but in [this](https://github.com/surrealdb/surrealkv/blob/137900b5f81e39f26b58c34c0c8bdddc8c42de58/src/storage/kv/entry.rs#L153) function, the timestamp is assigned externally, so think it would get a bit tricky because then I'll have to not depend on the entry ts. So thought of keeping it as it is for now.

PR #70, opened 2024-08-19
Implement set_at_ts

gsserge:
By default SKV sets the records' timestamp locally at commit time. Make it possible to override the timestamp using `set_at_ts`. This change is needed for CREATE with VERSION in SurrealDB.

- Implement `set_at_ts()`.
- Remove `commit_ts` and do not pass it around.
- Fix the signature of `get_at_ts()`.
- Bump the version.

### DIFF
--- a/src/storage/kv/store.rs
+++ b/src/storage/kv/store.rs
@@ -278,8 +278,6 @@ pub struct Task {
     done: Option<Sender<Result<()>>>,
     /// Transaction ID
     tx_id: u64,
-    /// Commit timestamp
-    commit_ts: u64,
     /// Durability
     durability: Durability,
 }
@@ -638,13 +636,7 @@ impl Core {
         let mut buf = BytesMut::new();
         let mut values_offsets = HashMap::with_capacity(req.entries.len());
 
-        encode_entries(
-            &req.entries,
-            req.tx_id,
-            req.commit_ts,
-            &mut buf,
-            &mut values_offsets,
-        );
+        encode_entries(&req.entries, req.tx_id, &mut buf, &mut values_offsets);
 
         let (segment_id, current_offset) = self.append_log(&buf, req.durability)?;
 
@@ -704,7 +696,7 @@ impl Core {
                 key: entry.key[..].into(),
                 value: index_value,
                 version: task.tx_id,
-                ts: task.commit_ts,
+                ts: entry.ts,
             });
         }
 
@@ -740,7 +732,6 @@ impl Core {
         &self,
         entries: Vec<Entry>,
         tx_id: u64,
-        commit_ts: u64,
         durability: Durability,
     ) -> Result<Receiver<Result<()>>> {
         let (tx, rx) = bounded(1);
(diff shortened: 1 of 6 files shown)


arriqaaq:
Hey @gsserge 

I was wondering if every record in a transaction can have a separate commit timestamp, or if all the records belonging within a transaction would have the same timestamp. If the latter, I was thinking if we can just accept timestamp variable on the commit method to set it.

If the former, I was wondering if it would adhere to this logic [here](https://github.com/surrealdb/vart/blob/d7d6fc52131853ef3656de6d4914cc47d4739268/src/art.rs#L1009)

gsserge:
I was thinking about having both - the individual write timestamp and the commit timestamp. I think we will need to support both cases:
1. users should be able to insert a bunch of records within one transaction but with different versions/timestamps
2. users should be able to bulk insert or create a bunch of records and then provide one version/timestamp at commit (next PR).

src/storage/kv/store.rs
```diff
@@ -700,11 +700,16 @@ impl Core {
 
             let index_value = encode_entry(entry);
 
+            let ts = if entry.ts == 0 {
```
arriqaaq, line 703: I was wondering if we need these conditional checks at all given that we are setting the commit ts for each entry

```
    fn assign_commit_ts(&mut self) -> u64 {
        let commit_ts = now();
        self.write_set.iter_mut().for_each(|entry| {
            if entry.ts == 0 {
                entry.ts = commit_ts;
            }
```

I think the commit_ts from task should be removed completely, just so simplify this so that we can just pass entry.ts everywhere. My bad I did not figure it out before. Do you think it is worth doing in this PR? Because else you have to add these checks everywhere

src/storage/kv/transaction.rs
```diff
@@ -160,6 +160,14 @@ impl Transaction {
         Ok(())
     }
 
+    /// Adds a key-value pair to the store with the given timestamp.
+    pub fn set_at_ts(&mut self, key: &[u8], value: &[u8], ts: u64) -> Result<()> {
```
arriqaaq, line 164: Can we also add one unit test for this to check if it passes with different varying timestamps, and maybe what happens when an older timestamp is passed (past the previous committed ts) to check if it fails?

gsserge: > when an older timestamp is passed (past the previous committed ts) to check if it fails?

Why do we want to prevent this behaviour? I think it should be fine to be able to create a version record with the timestamp older that the existing one.

arriqaaq: I mean with a timestamp older than the last transaction committed timestamp. Because [this](https://github.com/surrealdb/vart/blob/d7d6fc52131853ef3656de6d4914cc47d4739268/src/art.rs#L1017) condition will fail right

arriqaaq: My bad, please ignore, got confused between version and ts

gsserge: No worries!

arriqaaq (approved):

PR #75, opened 2024-09-11
Implement savepoints

gsserge:
- The API is modelled after RocksDB. `set_savepoint` is used to specify the state of a transaction to rollback to. `rollback_to_savepoint` is used to rollback to the latest savepoint.
- The savepoints are stackable. It's possible to set multiple savepoints within the same transaction and then rollback to each of them with the corresponding number of calls to `rollback_to_savepoint`.
- The implementation is quite simple. Each transaction stores the number of savepoints as `u32` with zero meaning none. When an entry is added to the transaction's write set, read set and scan set, its savepoint number is set to the current number of the transaction. In other words the entries are marked with the number of current savepoint. A call to `rollback_to_savepoint` merely removes all entries in the write set whose sevapoint number is equal to the transaction's one, and then it decrements it, stopping at 0.
- Bump the version.
- The draft SurrealDB PR is https://github.com/surrealdb/surrealdb/pull/4749.

### DIFF
--- a/src/storage/kv/transaction.rs
+++ b/src/storage/kv/transaction.rs
@@ -77,6 +77,53 @@ pub enum Durability {
     Immediate,
 }
 
+pub(crate) struct WriteSetEntry {
+    pub(crate) e: Entry,
+    pub(crate) savepoint_no: u32,
+}
+
+impl WriteSetEntry {
+    pub(crate) fn new(e: Entry, savepoint_no: u32) -> Self {
+        Self { e, savepoint_no }
+    }
+}
+
+pub(crate) struct ReadSetEntry {
+    pub(crate) key: Bytes,
+    pub(crate) ts: u64,
+    pub(crate) savepoint_no: u32,
+}
+
+impl ReadSetEntry {
+    pub(crate) fn new(key: Bytes, ts: u64, savepoint_no: u32) -> Self {
+        Self {
+            key,
+            ts,
+            savepoint_no,
+        }
+    }
+}
+
+pub(crate) struct ReadScanEntry {
+    pub(crate) start: Bound<VariableSizeKey>,
+    pub(crate) end: Bound<VariableSizeKey>,
+    pub(crate) savepoint_no: u32,
+}
+
+impl ReadScanEntry {
+    pub(crate) fn new(
+        start: Bound<VariableSizeKey>,
+        end: Bound<VariableSizeKey>,
+        savepoint_no: u32,
+    ) -> Self {
+        Self {
+            start,
+            end,
+            savepoint_no,
+        }
+    }
+}
+
 /// `Transaction` is a struct representing a transaction in a database.
 pub struct Transaction {
     /// `read_ts` is the read timestamp of the transaction. This is the time at which the transaction started.
@@ -99,13 +146,13 @@ pub struct Transaction {
 
     /// `write_set` is a vector of  entries.
     /// These are the changes that the transaction intends to make to the data.
-    pub(crate) write_set: Vec<Entry>,
+    pub(crate) write_set: Vec<WriteSetEntry>,
 
     /// `read_set` is the keys that are read in the transaction from the snapshot. This is used for conflict detection.
-    pub(crate) read_set: Mutex<Vec<(Bytes, u64)>>,
+    pub(crate) read_set: Mutex<Vec<ReadSetEntry>>,
 
     /// `read_key_ranges` is the key ranges that are read in the transaction from the snapshot. This is used for conflict detection.
-    pub(crate) read_key_ranges: Mutex<Vec<(Bound<VariableSizeKey>, Bound<VariableSizeKey>)>>,
+    pub(crate) read_key_ranges: Mutex<Vec<ReadScanEntry>>,
 
     /// `committed_values_offsets` is the offsets of values in the transaction post commit to the transaction log. This is used to locate the data in the transaction log.
     committed_values_offsets: HashMap<Bytes, usize>,
@@ -115,6 +162,9 @@ pub struct Transaction {
 
     /// `closed` indicates if the transaction is closed. A closed transaction cannot make any more changes to the data.
     closed: bool,
+
+    /// `savepoints` indicates the current number of stacked savepoints; zero means none.
+    savepoints: u32,
 }
 
 impl Transaction {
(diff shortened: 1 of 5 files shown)


src/storage/kv/entry.rs
```diff
@@ -31,6 +32,7 @@ impl Entry {
             metadata: None,
             value: Bytes::copy_from_slice(value),
             ts: 0,
+            savepoint_no: 0,
```
arriqaaq, line 35: Can we please include this in the write_set inside a transaction, making it a Vec<{Entry, savepoint_no}>?

The reason is that an entry is used for translation of records across multiple places (store, compaction etc) and this savepoint no is a property which is just relevant to the transaction internals. It will help avoid confusion later on in the entry.rs internals

src/storage/kv/transaction.rs
```diff
@@ -459,6 +466,73 @@ impl Transaction {
...
+        }
+        // Check that the savepoint is set
+        if self.savepoints == 0 {
+            return Err(Error::TransactionWithoutSavepoint);
+        }
+
+        // Remove entries marked for rollback since the last call
+        // to set_savepoint() from the write set, the write order map
+        // and the snapshot.
+        self.write_set.retain(|e| {
```
arriqaaq, line 514: Just thinking of the scenario where a scan was done on a range, and then rollbacked, I think we would need to track the read_key_ranges with savepoints too

src/storage/kv/transaction.rs
```diff
@@ -1997,4 +2071,132 @@ mod tests {
...
+        // Start another transaction and check again for the keys.
+        let mut txn2 = store.begin().unwrap();
+        assert_eq!(txn2.get(&key1).unwrap().unwrap(), value1);
+        assert!(txn2.get(&key2).unwrap().is_none());
+        assert!(txn2.get(&key3).unwrap().is_none());
+        txn2.commit().await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn multiple_savepoints_with_ssi() {
```
arriqaaq, line 2140: Could we also add a test for scan to check for rollback behaviour?

src/storage/kv/transaction.rs
arriqaaq: Would the read_set not required to be rollbacked too? Say I read a few keys since the last savepoint, now those keys in the read_set are used in conflict_detection. Wouldn't rolling back to a particular savepoint imply removing the tracked keys post the savepoint?

gsserge: Yes, you're right, I think we should track and clean up read set too on rollback.

src/storage/kv/entry.rs
gsserge: Sure, I actually implemented it as a vec of tuples first, but then decided to embed the savepoint no into Entry to save little bit on allocations when committing. But yeah, I guess in this case a small win in perf is not really worth it.

arriqaaq: Thank you!

src/storage/kv/transaction.rs
```diff
@@ -1997,4 +2136,237 @@ mod tests {
...
+        // Start another transaction and check again for the keys.
+        let mut txn2 = store.begin().unwrap();
+        assert_eq!(txn2.get(&key1).unwrap().unwrap(), value1);
+        assert!(txn2.get(&key2).unwrap().is_none());
+        assert!(txn2.get(&key3).unwrap().is_none());
+        txn2.commit().await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn savepont_with_concurred_read_txn_without_ssi() {
```
emmanuel-keller, line 2269:
```suggestion
    async fn savepoint_with_concurred_read_txn_without_ssi() {
```

src/storage/kv/transaction.rs
```diff
@@ -1997,4 +2136,237 @@ mod tests {
...
+        // Commit the transaction.
+        update_txn.commit().await.unwrap();
+        // Read transaction should commit without conflict after
+        // rolling back to the savepoint before reading `key`.
+        read_txn.rollback_to_savepoint().unwrap();
+        read_txn.commit().await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn savepont_with_concurred_read_txn_with_ssi() {
```
emmanuel-keller, line 2312:
```suggestion
    async fn savepoint_with_concurred_read_txn_with_ssi() {
```

src/storage/kv/transaction.rs
```diff
@@ -1997,4 +2136,237 @@ mod tests {
...
+        read_txn.get(&key).unwrap().unwrap();
+        update_txn.set(&key, &updated_value).unwrap();
+
+        update_txn.commit().await.unwrap();
+        read_txn.rollback_to_savepoint().unwrap();
+        read_txn.commit().await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn savepont_with_concurred_scan_txn() {
```
emmanuel-keller, line 2342:
```suggestion
    async fn savepoint_with_concurred_scan_txn() {
```

emmanuel-keller (approved):
LGTM

src/storage/kv/transaction.rs
```diff
@@ -459,6 +518,84 @@ impl Transaction {
...
+        }
+
+        // Remove entries marked for rollback since the last call
+        // to set_savepoint() from the write set, the write order map
+        // and the snapshot.
+        self.write_set.retain(|entry| {
+            if entry.savepoint_no == self.savepoints {
+                let hashed_key = sha256(entry.e.key.clone());
+                self.write_order_map.remove(&hashed_key);
+                self.snapshot
```
arriqaaq, line 570: There could be a potential issue here when the same key is overwritten. Let's say we have

```
    #[tokio::test]
    async fn foo() {
        // Scan set is only considered for SSI.
        let (store, _) = create_store(true);

        let k1 = Bytes::from("k1");
        let value = Bytes::from("test_value1");
        let value2 = Bytes::from("test_value2");

        let mut txn1 = store.begin().unwrap();
        txn1.set(&k1, &value).unwrap();
        txn1.set_savepoint().unwrap();
        txn1.set(&k1, &value2).unwrap();
        txn1.rollback_to_savepoint().unwrap();
        let val = txn1.get(&k1).unwrap();
        assert!(val.is_some());
    }
```

Because we are deleting the entire key from the snapshot (which does not track the savepoint number), this value is deleted, and could become an issue. The same would apply for scans too because scans also read from the snapshot

Issue #76, opened 2024-09-12
Savepoint rollback on an updated key should roll it back to previous version


arriqaaq:
There could be a potential issue here when the same key is overwritten. Let's say we have

```
    #[tokio::test]
    async fn savepoint_rollback_on_updated_key_should_pass() {
        // Scan set is only considered for SSI.
        let (store, _) = create_store(true);

        let k1 = Bytes::from("k1");
        let value = Bytes::from("test_value1");
        let value2 = Bytes::from("test_value2");

        let mut txn1 = store.begin().unwrap();
        txn1.set(&k1, &value).unwrap();
        txn1.set_savepoint().unwrap();
        txn1.set(&k1, &value2).unwrap();
        txn1.rollback_to_savepoint().unwrap();
        let val = txn1.get(&k1).unwrap();
        assert!(val.is_some());
    }
```

Because we are deleting the entire key from the snapshot (which does not track the savepoint number), this value is deleted, and could become an issue. The same would apply for scans too because scans also read from the snapshot

arriqaaq:
One way to solve this is to modify the set function on snapshot

    /// Set a key-value pair into the snapshot.
    pub fn set(&mut self, key: &VariableSizeKey, value: Bytes) {
        // TODO: need to fix this to avoid cloning the key
        // This happens because the VariableSizeKey transfrom from
        // a &[u8] does not terminate the key with a null byte.
        let key = &key.terminate();
        self.snap.insert(key, value, self.snap.version());
    }

to use save points as timestamp during insert than self.snap.version, and then add method on vart (remove_at_ts or something similar) which will not delete the entire key/node from the tree but just the latest value

PR #77, opened 2024-09-13
Savepoint rollback on an updated key should roll it back to previous version

gsserge:
Fixes https://github.com/surrealdb/surrealkv/issues/76.

- Make `write_set` a map from keys to a vec of value versions, each for a different savepoint. That way we could keep the values across savepoints and properly rollback to a previous version by re-adding it to the snapshot.
- ~~Making `write_set` a map also allowed me to get rid of `write_order_map` which has not been used anyway, meaning that we do not care about the order.~~ With it, we can also get rid of the expensive `sha256` calculations for keys, and also remove `sha2` as a dependency.
(Update) The order of writes within a transaction is now preserved with per-transaction sequence numbers.
- In `Transaction::get`, the write set is now checked first, before attempting the snapshot.
- Clean up `Transaction` by removing unused `buf` and `committed_values_offsets` fields.

### DIFF
--- a/src/storage/kv/transaction.rs
+++ b/src/storage/kv/transaction.rs
@@ -2,7 +2,7 @@ use std::ops::{Bound, RangeBounds};
 use std::sync::Arc;
 
 use ahash::{HashMap, HashMapExt};
-use bytes::{Bytes, BytesMut};
+use bytes::Bytes;
 use parking_lot::{Mutex, RwLock};
 use std::collections::hash_map::Entry as HashEntry;
 use vart::{art::QueryType, VariableSizeKey};
@@ -12,7 +12,7 @@ use crate::storage::kv::{
     error::{Error, Result},
     snapshot::{FilterFn, Snapshot, FILTERS},
     store::Core,
-    util::{convert_range_bounds, now, sha256},
+    util::{convert_range_bounds, now},
 };
 
 /// `Mode` is an enumeration representing the different modes a transaction can have in an MVCC (Multi-Version Concurrency Control) system.
@@ -79,12 +79,17 @@ pub enum Durability {
 
 pub(crate) struct WriteSetEntry {
     pub(crate) e: Entry,
-    pub(crate) savepoint_no: u32,
+    savepoint_no: u32,
+    seqno: u32,
 }
 
 impl WriteSetEntry {
-    pub(crate) fn new(e: Entry, savepoint_no: u32) -> Self {
-        Self { e, savepoint_no }
+    pub(crate) fn new(e: Entry, savepoint_no: u32, seqno: u32) -> Self {
+        Self {
+            e,
+            savepoint_no,
+            seqno,
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/storage/kv/transaction.rs
```diff
@@ -132,31 +132,25 @@ pub struct Transaction {
...
-    /// `write_order_map` is a mapping from sha256 of keys to their order in the write_set.
-    pub(crate) write_order_map: HashMap<Bytes, u32>,
-
-    /// `write_set` is a vector of  entries.
+    /// `write_set` is a map of keys to entries.
     /// These are the changes that the transaction intends to make to the data.
-    pub(crate) write_set: Vec<WriteSetEntry>,
+    /// The entries vec is used to keep different values for the same key for
+    /// savepoints and rollbacks.
+    pub(crate) write_set: HashMap<Bytes, Vec<WriteSetEntry>>,
```
arriqaaq, line 146: The intention of write_set being a vector was to preserve the order of writes

I am curious how this preserves the order in which writes are being written to the transaction. Let's say we have the following:

```
txn.set("k1", 1)
txn.set("k2", 2)
```

Could you help understand how would there be a guarantee k2 will be inserted after k1 here? The order will be important for deletes too if I think on the top of my head

gsserge: I'm actually not sure that the global order in needed. Let me also think about this over the weekend, but at the moment I do not think this is relevant, given that
- order per key is still preserved because of the vector, meaning that inserts and deletes for the same key are properly ordered
- global order between transactions is preserved with `commit_ts`, while in the same transactions all writes will have the same `commit_ts`.

arriqaaq: The sole reason to maintain order during initial implementation was for audit trails, to preserve the exact sequence of data insertion and the order of dependent operations, similar to a changelog. You are right, in the sense that not all types of applications might need this type of strict ordering, but I feel keeping the order preserved will be helpful in the long run. It is a safer bet, mainly because I can't think of all the scenarios right now :) but happy to discuss soon

gsserge: Let's extend the example above by adding another write for `k1`:
```
txn.set("k1", 1)
txn.set("k2", 2)
txn.set("k1", 3)
```
Now, with `write_order_map`, the write order will be [("k1", 3), ("k2", 2)], which is not the real time order that you'd expect, I think, because ("k1", 3) should come last. This happens because `write_order_map` remembers the _first_ time a key is written, and then it never changes the position even with additional writes later for the same key. I'm not sure that this property of ordering by the first write is particularly useful for auditing and thus worth preserving.

arriqaaq: You are right, this is a bug which needs fixing. Thanks for catching this. 

We will need to discuss and decide whether within a transaction we need to preserve this ordering or not, I see [badger](https://github.com/dgraph-io/badger/blob/499b1424621cc5730dab2315e3958643cb7131bc/txn.go#L261) also maintains a map, so they aren't doing it. In rocksdb, they seem to be just [appending](https://github.com/facebook/rocksdb/blob/0e04ef1a96d75acdcf2cf1802692375d62aea64c/db/write_batch.cc#L839) but I haven't looked too much into its details.

So we will need to decide if:

1) We get rid of ordering within a transaction
2) All three ( (k1, 1) (k2, 2) (k1, 3) ) get written
3) Any update to the previous written key updates the previous value ( (k1, 3) (k2, 2) )
4) Any update to the previous written key updates the last value ( (k2, 2) (k1, 3) )

gsserge: The write order within a transaction is now preserved using the write sequence numbering.

arriqaaq (approved):
LGTM!

Issue #80, opened 2024-09-18
Tokio reliance


stevefan1999-personal:
At the moment, these are the only few things that indicates Tokio requirement

https://github.com/surrealdb/surrealkv/blob/c6eadd17f8255f0fe6b0346fadf8925ad05d90e1/src/storage/kv/store.rs#L9-L12

While AsyncMutex should be able to be replaced with https://docs.rs/async-mutex/latest/async_mutex/index.html, for the task runner and async drop it would be a little more tricky. 

For task runner, we need to inverse the runtime control back to the user, so the user now has to implement a trait which will supply the async future to a user specific runtime spawner.

And for the async drop, we need to wait for https://rust-lang.github.io/async-fundamentals-initiative/roadmap/async_drop.html.

stevefan1999-personal:
I'm able to confirm that this code would allow SurrealKV to run on smol without extra compatibility shim after merging the two PRs!

```rust
use std::{env, sync::Arc};

use derive_more::derive::{Display, Error, From};
use surrealkv::{IsolationLevel, Options, Store};

use std::{
    future::Future,
    pin::{pin, Pin},
    result::Result,
    task::{Context, Poll},
};

use surrealkv::async_runtime::{JoinHandle, TaskSpawner};

#[derive(Debug, Display, Error, From)]
enum Error {
    SurrealKv(surrealkv::Error),
    Utf8(std::string::FromUtf8Error),
    ParseInt(std::num::ParseIntError),
    Io(std::io::Error),
}

fn main() -> Result<(), Error> {
    smol::block_on(async {
        let store = Store::with_spawner(
            Options {
                dir: env::current_dir()?.join("./foo"),
                isolation_level: IsolationLevel::SerializableSnapshotIsolation,
                ..Options::default()
            },
            Arc::new(SmolSpawner),
        )?;
        let mut tx = store.begin()?;
        if let Some(foo) = tx.get(b"foo")? {
            let foo = String::from_utf8(foo)?;
            println!("old value {}", foo);
            let mut foo: u32 = foo.parse()?;
            foo += 1;
            println!("new value {}", foo);
            tx.set(b"foo", foo.to_string().as_bytes())?;
        } else {
            tx.set(b"foo", b"1")?;
        }

        tx.commit().await?;
        Ok(())
    })
}

#[derive(Copy, Clone, Default)]
pub struct SmolSpawner;

impl TaskSpawner for SmolSpawner {
[...]

PR #81, opened 2024-09-18
remove tokio sync primitives

stevefan1999-personal:
As per https://github.com/surrealdb/surrealkv/issues/80, this PR will attempt to remove Tokio related sync primitives with `async_mutex` and `async_semaphore`. 

Technically speaking, the behavior shouldn't be that much differed, the few things that needs to be taken noticed on is only in fairness guarantee, whether one mutex lock would starve another, and whether or not two mutex lock re-entrancy would cause a live-lock or not.

As for `async_semaphore`, I noticed how the "write lock" for the oracle doesn't need a lock at all; it just need a critical section to prevent concurrent access. While I couldn't find any async critical section crate, I think replacing it with a semaphore would better indicate its functionality, hence the change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,10 +79,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1ca33f4bc4ed1babef42cad36cc1f51fa88be00420404e5b1e80ab1b18f7678c"
 dependencies = [
  "concurrent-queue",
- "event-listener",
+ "event-listener 4.0.3",
  "event-listener-strategy",
  "futures-core",
  "pin-project-lite",
+]
+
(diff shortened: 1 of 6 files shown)


arriqaaq:
Hey @stevefan1999-personal 

Sorry for the late reply, and thank you for these suggestions on surrealkv. I think they would be useful in the long run. Perhaps we can add them in a future milestone once everything is stable and working as expected, as we do not want to introduce any large breaking change at this time. I'll keep them open for now and maybe add it for a future milestone

gsserge:
The engine is now fully sync and does not rely on an async runtime https://github.com/surrealdb/surrealkv/pull/166.

PR #85, opened 2024-09-25
Read set and key ranges do not need mutex

gsserge:
Using a mutex seems to be too big of a price for `Transaction::get` with immutable `&self`.  We are already using mutable transactions everywhere, so there is no loss of functionality here.

Tested with SurrealDB.

### DIFF
--- a/benches/store_bench.rs
+++ b/benches/store_bench.rs
@@ -103,7 +103,7 @@ fn sequential_insert_read(c: &mut Criterion) {
 
                 let current_count = count.load(Relaxed);
                 if current_count <= max_count.load(Relaxed) {
-                    let txn = db.begin().unwrap();
+                    let mut txn = db.begin().unwrap();
                     txn.get(&current_count.to_be_bytes()[..]).unwrap();
                 }
             })
--- a/src/storage/kv/compaction.rs
+++ b/src/storage/kv/compaction.rs
@@ -512,7 +512,7 @@ mod tests {
         // Read the keys to the store
         for key in keys.iter().skip(num_keys_to_delete) {
             // Start a new read transaction
-            let txn = store.begin().unwrap();
+            let mut txn = store.begin().unwrap();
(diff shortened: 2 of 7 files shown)


arriqaaq:
Thanks @gsserge 

Is there any way to check how much of a difference it would make? Naturally, it doesn't make sense to have a transaction declared as mut for a get/scan operation. But if there is a significant bump, it would make sense to get rid of it.

I just ran a basic benchmark, the improvements were minor, but again, it simplifies the actual scenario a lot.

```
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use parking_lot::Mutex;

fn insert_into_array(arr: &mut Vec<i32>, value: i32) {
    arr.push(value);
}

fn insert_into_mutex_array(arr: &Mutex<Vec<i32>>, value: i32) {
    let mut arr = arr.lock();
    arr.push(value);
}

fn benchmark_insert(c: &mut Criterion) {
    let mut group = c.benchmark_group("insert_benchmark");

    group.bench_function("insert_into_array", |b| {
        let mut arr = Vec::new();
        b.iter(|| {
            insert_into_array(black_box(&mut arr), black_box(42));
        });
    });

    group.bench_function("insert_into_mutex_array", |b| {
        let arr = Mutex::new(Vec::new());
        b.iter(|| {
            insert_into_mutex_array(black_box(&arr), black_box(42));
        });
    });

    group.finish();
}

criterion_group!(benches, benchmark_insert);
criterion_main!(benches);
```

```
insert_benchmark/insert_into_array
                        time:   [2.5231 ns 2.5610 ns 2.5999 ns]

insert_benchmark/insert_into_mutex_array
                        time:   [5.4577 ns 5.4859 ns 5.5162 ns]
```

gsserge:
It's not about the perf per se, but more about the fact that with these two mutexes we do this weird interior mutability thing that does not make much sense to me.

arriqaaq (approved):

PR #86, opened 2024-09-25
Snapshots do not need start_ts

gsserge:
`Snapshot::start_ts` is confusing because it can lead to using `now()` for getting the value which is clearly wrong. 
It's actually not used at all, since the snapshot object implicitly records the version/timestamp as the largest version stored in the root node.

Tested with SurrealDB.


arriqaaq (approved):
I agree! LGTM

PR #92, opened 2024-10-14
Do not encode index entries in memory

gsserge:
Use explicit `enum IndexValue` with the `DiskIndexEntry` and `MemIndexEntry` variants, instead of encoding index values in memory with `ValueRef`.

As a further optimisation, we could even make `struct Indexer` generic and parameterise it with either `DiskIndexEntry` or `MemIndexEntry` when creating a Store.

Tested with SurrealDB with `cargo make ci-api-integration-surrealkv`.


src/storage/kv/indexer.rs
```diff
@@ -50,3 +50,82 @@ impl Indexer {
...
+}
+
+#[derive(Clone)]
+pub struct MemIndexEntry {
+    metadata: Option<Metadata>,
+    value: Bytes,
+}
+
+#[derive(Clone)]
+pub enum IndexValue {
```
arriqaaq, line 70: Given that you are getting rid of the `Value` trait, and all the functions for IndexValue are private (and no one can call any function over it)

1) We make this private to the crate, and consequently make the entire Snapshot and its functions private to the crate

2) We keep the Value trait (removing unnecessary functions like ts etc) and implement it on IndexValue 

I am leaning towards 1, we can make Snapshot public later if needed. LMKWYT

src/storage/kv/transaction.rs
```diff
@@ -732,20 +729,20 @@ impl Transaction {
             }
 
             // Create a new value reference and decode the value.
-            let mut val_ref = ValueRef::new(self.core.clone());
+            // let mut val_ref = ValueRef::new(self.core.clone());
```
arriqaaq, line 732: Leftover? Probably this entire section needs to be removed?

src/storage/kv/transaction.rs
```diff
@@ -786,7 +783,7 @@ impl Transaction {
         &self,
         key: &VariableSizeKey,
         query_type: QueryType,
-    ) -> Result<(Bytes, u64, u64)> {
+    ) -> Result<(IndexValue, u64, u64)> {
```
arriqaaq, line 786: If this is not returning Bytes, and IndexValue has no method to be accessed publicly, wouldn't this create a problem on surrealdb? Since indexvalue is private, I'd expect the return type to remain the same

src/storage/kv/indexer.rs
gsserge: I'm also fine with the Option 1.

gsserge: `Snapshot` and `IndexValue` are now private to the crate.

src/storage/kv/transaction.rs
gsserge: Removed.

src/storage/kv/transaction.rs
gsserge: The actual value is now being resolved.

arriqaaq (approved):

PR #94, opened 2024-10-17
Feat/add scan all versions api

arriqaaq:
## Description

This PR adds a `scan_all_versions` method to scan all the versions of the keys within the specified range.

### DIFF
--- a/src/storage/kv/transaction.rs
+++ b/src/storage/kv/transaction.rs
@@ -628,6 +628,7 @@ impl Transaction {
 }
 
 /// Implement Versioned APIs for read-only transactions.
+/// These APIs do not take part in conflict detection.
 impl Transaction {
     fn ensure_read_only_transaction(&self) -> Result<()> {
         // If the transaction is closed, return an error.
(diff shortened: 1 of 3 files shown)


src/storage/kv/transaction.rs
```diff
@@ -796,6 +800,55 @@ impl Transaction {
...
+
+        // Initialize an empty vector to store the results.
+        let mut results = Vec::new();
+
+        // Get a range iterator for the specified range.
+        let snap = self.snapshot.as_ref().unwrap().read();
+        let ranger = snap.range_with_versions(range);
+
+        // Iterate over the keys in the range.
+        for (key, value, _, ts) in ranger {
```
gsserge, line 824: If `key` is  declared as `mut`, then there's no need for `let mut key = key;` later:
```rust
        for (mut key, value, _, ts) in ranger {
```

src/storage/kv/transaction.rs
```diff
@@ -796,6 +800,55 @@ impl Transaction {
...
+            // If a limit is set and we've already got enough results, break the loop.
+            if let Some(limit) = limit {
+                if results.len() >= limit {
+                    break;
+                }
+            }
+
+            // Determine if the record is soft deleted based on the metadata.
+            let mut is_deleted = false;
```
gsserge, line 833: This can be simplified with the `Option::is_some_and` predicate
```rust
let is_deleted = value.metadata().is_some_and(|md| md.is_tombstone());
```

gsserge (approved):

src/storage/kv/transaction.rs
arriqaaq: Thank you, fixed it

src/storage/kv/transaction.rs
arriqaaq: Learnt something new! Fixed

AlbertMarashi:
does this bring something like this proposal closer to fruition
https://github.com/surrealdb/surrealdb/issues/4860

arriqaaq:
Hey @AlbertMarashi 

This fixes the export version bug you had reported, where all the versions were not exported for a record

PR #102, opened 2024-10-24
fix: sdb bug on hard delete and reinsert

arriqaaq:
## Description

Fixes a bug where scan after a hard_delete would create a read conflict. This is because snapshots created were always based on the vart version, which rolls back (or decreases) when a hard_delete occurs.

### DIFF
--- a/src/storage/kv/snapshot.rs
+++ b/src/storage/kv/snapshot.rs
@@ -13,13 +13,19 @@ pub(crate) const FILTERS: [fn(&IndexValue) -> Result<()>; 1] = [ignore_deleted];
 /// A versioned snapshot for snapshot isolation.
 pub(crate) struct Snapshot {
     snap: VartSnapshot<VariableSizeKey, IndexValue>,
+    ts: u64,
 }
 
 impl Snapshot {
     pub(crate) fn take(store: Arc<Core>) -> Result<Self> {
-        let snapshot = store.indexer.write().snapshot();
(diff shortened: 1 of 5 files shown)


src/storage/kv/snapshot.rs
```diff
@@ -13,13 +13,19 @@ pub(crate) const FILTERS: [fn(&IndexValue) -> Result<()>; 1] = [ignore_deleted];
 /// A versioned snapshot for snapshot isolation.
 pub(crate) struct Snapshot {
     snap: VartSnapshot<VariableSizeKey, IndexValue>,
+    ts: u64,
```
gsserge, line 16: Can we name this `version`?

arriqaaq: It is actually the timestamp. Because the version is set on the snapshot and internally it remains static, and what we are passing here is the ts as seen [here](https://github.com/surrealdb/vart/blob/2f33a89d7d890602c6f2a4d2c23f4a7199cb168a/src/snapshot.rs#L24)

arriqaaq: Actually let me get rid of that altogether and make every insert inside the snapshot have a ts of 0 as it hardly matters

arriqaaq: I've removed this, rather adding now as ts for every insert in snapshot

gsserge (approved):

Issue #108, opened 2024-10-25
Add tests/checks in surrealkv to assert and confirm that transactions ids are always moving forward


arriqaaq:
This check was previously in bulk_insert inside vart. We need to add the same logic inside surrealkv now that it manages checking on versions

PR #110, opened 2024-10-28
chore: add test for concurrent inserts

arriqaaq:
## Description

This PR adds the following:

1) Adds check if the snapshot is not lesser than index version
2) Modifies Transaction::commit to return (tx_id, commit_ts) for verification
3) Add basic test to verify if concurrent transactions committed have incremental txn ids

Addresses https://github.com/surrealdb/surrealkv/issues/108

### DIFF
--- a/src/storage/kv/snapshot.rs
+++ b/src/storage/kv/snapshot.rs
@@ -19,11 +19,26 @@ pub(crate) struct Snapshot {
 
 impl Snapshot {
     pub(crate) fn take(store: &Core) -> Result<Self> {
-        // Each snapshot is created at a version that is one greater than the current version.
+        // Acquire a read lock on the indexer to get the current index version.
         let index = store.indexer.read();
-        let version = store.read_ts() + 1;
+        let index_version = index.version();
+
+        // Calculate the snapshot version as one greater than the current read timestamp.
+        let read_version = store.read_ts() + 1;
+
+        // Ensure that the snapshot read version is not older than the current index version.
+        if read_version < index_version {
+            return Err(Error::SnapshotVersionIsOld(read_version, index_version));
+        }
+
+        // Clone the current index to create the snapshot.
         let snap = index.index.clone();
-        Ok(Self { snap, version })
+
+        // Return the new snapshot with the calculated version.
+        Ok(Self {
+            snap,
+            version: read_version,
+        })
     }
 
     /// Set a key-value pair into the snapshot.
--- a/src/storage/kv/transaction.rs
+++ b/src/storage/kv/transaction.rs
@@ -3378,14 +3378,19 @@ mod tests {
         }
     }
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread")]
     async fn test_concurrent_transactions() {
         let (store, _) = create_store(false);
         let store = Arc::new(store);
 
+        // Define the number of concurrent transactions
+        let num_transactions = 1000;
+
         // Define key-value pairs for the test
-        let keys: Vec<Bytes> = (0..10).map(|i| Bytes::from(format!("key{}", i))).collect();
-        let values: Vec<Bytes> = (0..10)
+        let keys: Vec<Bytes> = (0..num_transactions)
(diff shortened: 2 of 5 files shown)


src/storage/kv/snapshot.rs
```diff
@@ -19,11 +19,26 @@ pub(crate) struct Snapshot {
...
+        // Acquire a read lock on the indexer to get the current index version.
         let index = store.indexer.read();
-        let version = store.read_ts() + 1;
+        let index_version = index.version();
+
+        // Calculate the snapshot version as one greater than the current read timestamp.
+        let read_version = store.read_ts() + 1;
+
+        // Ensure that the snapshot read version is not older than the current index version.
+        if read_version < index_version {
```
arriqaaq, line 30: @gsserge does it make sense to add this check?

arriqaaq: There is this race test which I wrote for previous version to check for errors, and it doesn't happen here, and I verified using this

```
    #[tokio::test(flavor = "multi_thread")]
    async fn test_version_snapshot_race() {
        let (store, _) = create_store(false);
        let store = Arc::new(store);
        
        // Flags to coordinate the timing of operations
        let commit_started = Arc::new(AtomicBool::new(false));
        let commit_finished = Arc::new(AtomicBool::new(false));
        
        // First transaction: will be in commit phase when second transaction starts
        let store_clone = Arc::clone(&store);
        let commit_started_flag = Arc::clone(&commit_started);
        let commit_finished_flag = Arc::clone(&commit_finished);
        
        let commit_handle = tokio::spawn(async move {
            let mut txn = store_clone.begin().await.unwrap();

            txn.set(&Bytes::from("key1"), &Bytes::from("value1")).unwrap();
            
            // Signal that we're about to commit
            commit_started_flag.store(true, Ordering::SeqCst);
            
            let start = std::time::Instant::now();
            tokio::time::sleep(Duration::from_millis(100)).await;
            txn.commit().await.unwrap();

            let mut txn = store_clone.begin().await.unwrap();
            txn.set(&Bytes::from("key2"), &Bytes::from("value1")).unwrap();
            txn.commit().await.unwrap();

            
            commit_finished_flag.store(true, Ordering::SeqCst);
        });
        
        // Second transaction: will try to read version and take snapshot after commit completes
        let store_clone = Arc::clone(&store);
        let commit_started_flag = Arc::clone(&commit_started);
        let commit_finished_flag = Arc::clone(&commit_finished);
        
        let race_handle = tokio::spawn(async move {
            // Wait for the first transaction to start committing
            while !commit_started_flag.load(Ordering::SeqCst) {
                // tokio::time::sleep(Duration::from_millis(10)).await;
                std::thread::sleep(Duration::from_millis(10));
            }
            
            // Now we know the first transaction is in its commit phase
            let txn = store_clone.begin_with_closure(|| async {
                // Wait for commit to finish before proceeding
                while !commit_finished_flag.load(Ordering::SeqCst) {
                    std::thread::sleep(Duration::from_millis(10));
                    // tokio::time::sleep(Duration::from_millis(10)).await;
                }
                
                // Now add delay after commit has finished
                std::thread::sleep(Duration::from_millis(1000));
            }).await.unwrap();
        });
        
        // Wait for both operations to complete
        commit_handle.await.unwrap();
        race_handle.await.unwrap();
        
        store.close().await.unwrap();
    }

```

src/storage/kv/transaction.rs
```diff
@@ -439,7 +439,7 @@ impl Transaction {
     }
 
     /// Commits the transaction, by writing all pending entries to the store.
-    pub async fn commit(&mut self) -> Result<()> {
+    pub async fn commit(&mut self) -> Result<Option<(u64, u64)>> {
```
gsserge, line 442: I'm not sure that exposing internal internal details like txid and commit_ts in a public API is a good idea.

arriqaaq: Noted, made it private to keep this api unchanged

gsserge (approved):

PR #118, opened 2024-11-05
chore: remove writes to snapshot

arriqaaq:


src/storage/kv/transaction.rs
```diff
@@ -542,7 +659,56 @@ impl Transaction {
...
+            .filter_map(|(key, entries)| {
+                if !Transaction::is_key_in_range(key, &range) {
+                    return None;
+                }
+                entries.last().map(|entry| (key, entry))
+            })
+            .collect();
+
+        // Sort by key
+        entries.sort_by(|a, b| a.0.cmp(b.0));
```
gsserge, line 687: It think it might actually be better to change the Write Set from `HashMap` to `BTreeMap`. That way you won't have to collect the latest values upfront in a vector and sort them, since `BTreeMap`'s iterator is ordered. Maybe this will be enough to keep the scan perf in par with the current one.

gsserge:
Handled in https://github.com/surrealdb/surrealkv/pull/151

PR #119, opened 2024-11-11
chore: add test for migration from old version

arriqaaq:


Cargo.toml
```diff
@@ -41,6 +41,7 @@ libc = "0.2.155"
 fmmap = "0.3.3"
 memmap2 = "0.9.4"
 num_cpus = "1.16.0"
+skv44 = { package = "surrealkv", version = "0.4.4" }
```
gsserge, line 44: Just to be safe, let's pin the [exact version](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#comparison-requirements) of skv with
```toml
skv44 = { package = "surrealkv", version = "=0.4.4" }
```

gsserge (approved):

Cargo.toml
arriqaaq: noted!

PR #123, opened 2024-11-13
Improve README

arriqaaq:
## Description

Add detailes of surrealkv to the README, including architecture, strengths/limitations, benchmarks etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![License](https://img.shields.io/badge/license-Apache_License_2.0-00bfff.svg?style=flat-square)](https://github.com/surrealdb/surrealkv)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,28 +1,458 @@
...
+let all_versions = tx.scan_all_versions(range, Some(10))?;
+```
+
+## Advanced Features
+
+### Durability Levels
+
+The `Durability` enum provides two levels of durability for transactions:
+
+- `Eventual`: Commits with this durability level are guaranteed to be persistent eventually. The data is written to the disk, but it is not fsynced before returning from `Transaction::commit`. This is the default durability level.
```
gsserge, line 170: The data is written to the kernel buffer, not disk.

gsserge (approved):

PR #131, opened 2024-11-18
fix: do not allow blind writes in SI

arriqaaq:
## Description

Snapshot isolation forbids write-write conflict. This PR fixes it as current SI allows blind writes


arriqaaq:
Closing this PR in favor of https://github.com/surrealdb/surrealkv/pull/132

PR #132, opened 2024-11-18
Feat/improve snapshot isolation

arriqaaq:
## Description

This PR makes changes to the Snapshot Isolation implementation (both SI and SSI). It keeps SI similar to the definition in [paper](https://arxiv.org/abs/2405.18393) and makes SSI stronger


src/storage/kv/oracle.rs
```diff
@@ -304,244 +204,115 @@ fn key_in_range(
 /// - `commit_tracker` maintains information about committed transactions and their timestamps.
 /// - `txn_mark` is a watermark used to block new transactions until previous commits are visible.
 /// - `read_mark` is another watermark that marks the visibility of read operations to other transactions.
-///
-/// The serializable snapshot isolation (SSI) algorithm implemented here is inspired from BadgerDB.
```
gsserge, line 308: Let's also change the comment and state that this is an implementation of Write Snapshot Isolation as described in https://arxiv.org/abs/2405.18393.

arriqaaq: Thank you, added the reference

gsserge (approved):

PR #137, opened 2024-11-22
Add CLI Tool for SurrealKV management and maintenance

arriqaaq:
## Description 

This PR introduces a command-line interface (CLI) for SurrealKV storage management and maintenance.

Key Features:
- Version Analysis: Identify keys with multiple versions
- Pruning: Remove old versions while keeping the latest (by transaction ID)
- Statistics: Get database metrics
- Repair: Fix corrupted commit logs

Documentation:
- Added detailed README with examples
- Command help text with clear option descriptions


gsserge:
I think from the layout perspective we can make things a bit simple since the same package can have both a bin crate and a lib crate, so there's no need to a workspace and a standalone `cli` binary crate. You could just put the CLI source into `bin/`.

gsserge (approved):

PR #142, opened 2024-11-27
Implement Transaction::keys_with_tombstones

gsserge:
Returns all existing keys within the specified range, including soft-deleted and thus hidden by tombstones.
This is needed for the `REMOVE WITH EXPUNGE` functionality to work properly in SurrealDB. Without this change soft-deleted keys and their values will not be removed.

Let me know if another name is a better choice for this method. Maybe `all_index_keys()`?

### DIFF
--- a/src/storage/kv/transaction.rs
+++ b/src/storage/kv/transaction.rs
@@ -456,6 +456,29 @@ impl Transaction {
         Ok(results)
     }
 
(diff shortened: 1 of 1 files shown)


arriqaaq:
LGTM! Name is very explicit and clear

arriqaaq (approved):

PR #158, opened 2025-01-07
feat: add keys api

arriqaaq:
## Description

This PR does the following

1. Add `keys` function with signature (&self, range, limit)
2. Modify keys_at_ts with a limit argument, with signature (&self, range, ts, limit)
3. Modify keys_with_tombstones function with a limit argument, with signature (&self, range, limit)

### DIFF
--- a/src/snapshot.rs
+++ b/src/snapshot.rs
@@ -178,9 +178,14 @@ mod tests {
             .expect("Failed to begin transaction");
 
         let range = "k1".as_bytes()..="k2".as_bytes();
-        let keys: Vec<_> = txn.keys_at_ts(range.clone(), ts).collect();
+        let keys: Vec<_> = txn.keys_at_ts(range.clone(), ts, None).collect();
         assert_eq!(keys[0], b"k1");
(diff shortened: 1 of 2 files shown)


src/snapshot.rs
```diff
@@ -131,11 +131,37 @@ impl Snapshot {
...
+        &'a self,
+        range: R,
+        ts: u64,
+        limit: Option<usize>,
+    ) -> Vec<&'a [u8]>
     where
         R: RangeBounds<VariableSizeKey> + 'a,
     {
-        self.snap.keys_at_ts(range, ts).collect()
+        let iter = self.snap.keys_at_ts(range, ts);
```
gsserge, line 143: Aren't we removing `keys_at_ts` in https://github.com/surrealdb/vart/pull/79?

arriqaaq: Yup, chucking this

arriqaaq: Removed! Just rewrote this PR

gsserge (approved):

PR #161, opened 2025-01-13
Make Scan APIs return Iterator

arriqaaq:
## Description

This PR does the following:

1. Updates vart and modifies changes required inside skv
2. Makes scan related APIs return iterator

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -397,7 +397,7 @@ impl Transaction {
         &'b mut self,
         range: R,
         limit: Option<usize>,
-    ) -> impl Iterator<Item = Result<ScanResult<'b>>> + 'b
+    ) -> impl Iterator<Item = Result<ScanResult<'b>>>
     where
         R: RangeBounds<&'b [u8]>,
     {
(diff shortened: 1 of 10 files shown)


src/transaction.rs
```diff
@@ -513,16 +431,17 @@ impl Transaction {
     /// Returns all existing keys within the specified range, including soft-deleted
     /// and thus hidden by tombstones.
     /// The returned keys are not added to the read set and will not cause read-write conflicts.
-    pub fn keys_with_tombstones<'b, R>(&'b self, range: R) -> Result<Vec<Vec<u8>>>
+    pub fn keys_with_tombstones<'b, R>(&'b self, range: R) -> impl Iterator<Item = &'b [u8]> + 'b
```
gsserge, line 434: I don't think ' + 'b' is needed here; same goes for another three instances in this file.

gsserge (approved):

src/transaction.rs
arriqaaq: Thanks for pointing it out

PR #166, opened 2025-01-16
Make it sync

gsserge:


src/store.rs
```diff
@@ -134,118 +75,37 @@ impl Store {
...
-
-    /// Compacts the store.
-    pub async fn compact(&self) -> Result<()> {
-        if let Some(inner) = self.inner.as_ref() {
-            inner.compact().await?;
+        if self.is_compacting.load(Ordering::SeqCst) {
+            return Err(Error::CompactionAlreadyInProgress);
         }
 
+        self.core.close()?;
```
arriqaaq, line 96: I was wondering if during close, we should take a lock on the oracle? This would guarantee there are no ongoing transactions at all right? And given that if we achieve a lock on it, this means there are no in-flight transactions?

gsserge: Taking the oracle lock would not prevent new transactions from starting, it would only block commit.

Now we only rely on a single lock, the active segment lock. I'd really like to not introduce any more explicit locking unless absolutely necessary, because using 2+ locks would open more possibilities for deadlocking.

arriqaaq: I understand it wouldn't prevent, but only block commit, but my thought was that whether we should use a high-level coordination lock for store closure (and hence the thought of the oracle lock). Currently, using just the active segment lock works well since we only need to coordinate log closure (we don't require index to close, or any other component). However, if we add more components that require coordinated shutdown (like btree, etc), relying on individual component locks might become confusing. A parent-level lock I think could provide cleaner orchestration for multi-component shutdown so that we don't have to revisit this in the future. 

Not a blocker now, we can definitely work on this when some other component comes in. 

Probably add a comment about how the close coordination works on the aol log lock to remember in future?

gsserge: Agree, but let's not do that prematurely - that's why I mentioned "unless absolutely necessary". It's not needed _now_, and that's important. Once and if we really needs this, we will add it.

gsserge: Added a comment in `Core::close()`.

arriqaaq: Awesome

src/oracle.rs
```diff
@@ -1,7 +1,8 @@
 use bytes::Bytes;
 use std::ops::{Bound, RangeBounds};
 use std::sync::atomic::{AtomicU64, Ordering};
-use tokio::sync::Mutex as AsyncMutex;
+// use tokio::sync::Mutex as AsyncMutex;
```
arriqaaq, line 4: Leftover

src/oracle.rs
```diff
@@ -1,7 +1,8 @@
...
+use std::sync::Mutex;
```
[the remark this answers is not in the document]
gsserge: AFAIK over the past few years the perf gap between std::sync and parking_lot has been mostly closed, at least on Linux.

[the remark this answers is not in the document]
arriqaaq: Because we are already using it in aol and in store, was wondering if we could use the same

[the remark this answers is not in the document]
arriqaaq: Ah ok, wasn't aware. Not a blocker but just that since parking_lot is already used in several places, we can either use one of them (either std mutex or parking_lot) to keep it uniform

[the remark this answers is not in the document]
gsserge: Seems like std Mutex performs better when highly contended which is our case with a large number of concurrent transactions. I've updated the PR to use parking_lot just for the sake of uniformity with other locks in the code base, but sometime later we might want to change every Mutex and RwLock to std and run the benches.

arriqaaq (approved):

src/oracle.rs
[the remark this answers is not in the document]
arriqaaq: Ack, I opened an issue to keep track of it

https://github.com/surrealdb/surrealkv/issues/167

PR #176, opened 2025-01-27
Improve read concurrency on aol 

arriqaaq:
## Description

This PR improves concurrent read handling in AOL. The writer has a separate file descriptor which is used for writing to the active segment, and the readers have separate file descriptor for reading.


src/store.rs
```diff
@@ -478,7 +478,7 @@ impl Core {
     }
 
     fn append_log(&self, tx_record: &BytesMut, durability: Durability) -> Result<(u64, u64)> {
-        let mut clog = self.clog.as_ref().unwrap().write();
+        let clog = self.clog.as_ref().unwrap().write();
```
gsserge, line 481: With `Aol` made thread safe using `Mutex` internally, I think we might be able do drop the `RwLock` protecting `clog' at the `Store` level, because it will serve as the serialisation mechanism between `close()` and `append_log()` as stated in https://github.com/surrealdb/surrealkv/pull/176/files#diff-41a991785a2eba02345db5fdcbf15b660f27b2841b8f61a084a7db66e69bf3eeL413-L421

gsserge (approved):

src/log/aol.rs
```diff
@@ -183,65 +187,49 @@ impl Aol {
...
-                r += bytes_read;
-            }
-            Err(e) => match e {
-                Error::Eof => {
-                    return Err(Error::Eof);
-                }
-                _ => return Err(e),
-            },
+        // Get segment from cache (works for both active and inactive segments)
+        let segment = self.segment_cache.get_or_insert_with(&segment_id, || {
```
gsserge, line 191: As a follow-up, maybe explore the possibility of keeping the active segment out of the cache for reads.

src/store.rs
arriqaaq: I agree, I thought of making/testing those changes in a separate PR so that any anomalies introduced through this code can then be isolated

src/log/aol.rs
arriqaaq: I just wanted to avoid having an RWLock around the active segment. My thought is that since write_all writes to the os page cache in the kernel space, if a new file descriptor is opened to the same active segment, it can still read the data as we are not buffering anything in user space. 

Do you think there could be any issues with this?

PR #186, opened 2025-02-14
Add support for reverse iter in scan

arriqaaq:


Cargo.toml
```diff
@@ -29,8 +29,9 @@ parking_lot = "0.12.1"
 lru = "0.12.0"
 bytes = "1.5.0"
 quick_cache = "0.6.0"
-vart = "0.9.1"
+vart = {git = "https://github.com/surrealdb/vart", branch = "backward-iters"}
```
gsserge, line 32: Needs a new `vart` release.

gsserge (approved):

Cargo.toml
arriqaaq: Updated

PR #195, opened 2025-03-20
Fix compaction on Windows by ensuring file handles are released before renaming

### LINKED ISSUE #194: SurrealKV Compaction Issue
### Issue Description

Compaction is not working as expected on Windows. Even manually triggering `store.compact().unwrap()` fails with a file permission error.

When attempting to run `store.compact().unwrap()` manually, the following error is returned:

```
called `Result::unwrap()` on an `Err` value: IoError(Os { code: 5, kind: PermissionDenied, message: "Access is denied." })
```

Despite running the program in an Administrator PowerShell session and ensuring the folder permissions allow full access, compaction does not reduce the disk size, and the error persists.

### Steps to Reproduce

1. Create a new SurrealKV store with the following options:

```rust
let mut opts = Options::new();
opts.dir = "./kv-store".into();
opts.enable_versions = false;
opts.max_value_cache_size = 25000;
opts.max_segment_size = 1 << 27;
opts.max_compaction_segment_size = 1 << 29;
let store = Store::new(opts)?;
```

2. Insert and delete several keys to create segments.
3. Call `store.compact().unwrap()`.
4. Observe the error:

```
called `Result::unwrap()` on an `Err` value: IoError(Os { code: 5, kind: PermissionDenied, message: "Access is denied." })
```

### Expected Behavior

Compaction should reduce the disk size and run without permission errors, especially when executed with administrative privileges.

### Actual Behavior

- Compaction does not occur.
- `store.compact()` returns an "Access is denied" error.
- The disk space usage grows continuously without reduction.

### Environment

- **SurrealKV Version:** Latest (0.9.1)
- **Operating System:** Windows 11
- **Rust Version:** rustc 1.83.0
- **PowerShell:** Running with Administrator privileges

### Additional Context

- I noticed that in one of the existing issues, it was mentioned:
  > "The Windows tests are currently ignored as the compaction tests in SKV fail because of file permission issue. This can be fixed later."
- Is this a known limitation on Windows, or is there a workaround to get compaction working?
- Any advice or fixes would be greatly appreciated!

Thank you!

raghav7203:
This PR fixes a compaction failure on Windows caused by open file handles preventing the rename operation. Explicitly dropping temp_writer and manifest before the rename ensures the resources are freed, allowing the compaction process to complete successfully. Additionally, verified that all compaction tests now pass on Windows.

Fixes #194

### DIFF
--- a/src/compaction.rs
+++ b/src/compaction.rs
@@ -196,6 +196,10 @@ impl Store {
 
         temp_writer.close()?;
 
+        // Explicitly drop temp_writer and manifest to release file handles before renaming
+        drop(temp_writer);
+        drop(manifest);
+
         // Finalize compaction by renaming the temporary directory
         fs::rename(tmp_merge_dir, merge_dir)?;
 


arriqaaq (approved):

arriqaaq:
Thank you and congratulations @raghav7203!

Issue #199, opened 2025-07-24
Cannot insert multiple versions of a key-value pair in the same transaction.


Jezza:
There's a bit of an optimisation that's preventing us from inserting multiple datapoints in one go.

On the latest version, it's here: `src/transaction.rs:357`

This is preventing us from inserting different versions.
We synchronise a lot of data (including version info) within a single transaction, which means that some of the data isn't copied over.

Is there a way we can optionally disable this?

Happy to contribute a PR to change this.

arriqaaq:
Hey @Jezza 

I assume you'd want to insert multiple versions for the same key? I think we could just have a separate `write_with_version` or something similar which doesn't do the check. Please feel free to raise a PR, if not, I'll have a look at it

Jezza:
Hey, just had a further look into it, and such a thing doesn't work:

(this is after I've made some changes locally already)

```rust
    let store = surrealkv::Store::new(surrealkv::Options {
        dir: PathBuf::from("./data"),
        disk_persistence: false,
        enable_versions: true,
        ..surrealkv::Options::default()
    }).unwrap();

    let mut tx = store.begin().unwrap();

    tx.set_at_ts(b"a", b"", 1).unwrap();
    tx.set_at_ts(b"a", b"", 2).unwrap();

    println!("Scanning");

    for result in tx.scan_all_versions(b"\x00".as_slice()..b"\xFF".as_slice(), None) {
        let (_, _, ts, _) = result.unwrap();
        println!("@ {}", ts);
    }

    tx.commit().unwrap();

    println!("Scanning");
    let mut tx = store.begin().unwrap();
    for result in tx.scan_all_versions(b"\x00".as_slice()..b"\xFF".as_slice(), None) {
        let (_, _, ts, _) = result.unwrap();
        println!("@ {}", ts);
    }

    tx.set_at_ts(b"a", b"", 1).unwrap();
    tx.set_at_ts(b"a", b"", 2).unwrap();

    println!("Scanning");
    for result in tx.scan_all_versions(b"\x00".as_slice()..b"\xFF".as_slice(), None) {
        let (_, _, ts, _) = result.unwrap();
        println!("@ {}", ts);
    }

    tx.commit().unwrap();

    let tx = store.begin_with_mode(surrealkv::Mode::ReadOnly).unwrap();

    for result in tx.scan_all_versions(b"\x00".as_slice()..b"\xFF".as_slice(), None) {
        let (_, _, ts, _) = result.unwrap();
[...]

PR #200, opened 2025-07-24
Only overwrite the previous entry if we share savepoint and the entry is replacing the previous.

Jezza:
I originally had it as:
`last_entry.savepoint_no == ws_entry.savepoint_no && last_entry.version == ws_entry.version`
But I realised that wouldn't work as if people wanted to use `0` as a timestamp (which is replaced with the current ts) this would still prevent that.

So the solution, just check if the entry itself can replace existing entries.


arriqaaq:
Hi @Jezza 

Thanks for the PR, appreciate it. I'd like the changes to be very minimal so that it is easy to track and the PR is focused. As discussed in https://github.com/surrealdb/surrealkv/issues/199#issuecomment-3113099399, I'd rather have a different write method specific to versions than modify the existing code.

You can always open multiple PRs for refactor or any other changes.

arriqaaq:
The VersionScanIterator is not meant to work with ongoing write transactions, but only for read-only transactions

arriqaaq:
Sorry for the delay @Jezza 

This PR looks good to me, can you just add a couple of unit tests to confirm it works as expected? I'll merge it post that

arriqaaq (approved):

PR #320, opened 2026-01-26
Correctly backoff when adding the commit to the queue.

Jezza:


arriqaaq (approved):
LGTM, thank you @Jezza

PR #380, opened 2026-04-27
fix(snapshot): stack overflow in backward iteration over tombstoned keys

### LINKED ISSUE #379: Stack overflow in backward range iteration over tombstoned keys
Hit this on a Tauri desktop app running an embedded surrealdb store on macOS. After a while the app started crashing on launch with `EXC_BAD_ACCESS` on the stack guard page. The crash report showed ~1000 levels of recursion alternating between `skip_to_valid_backward` and `find_latest_visible_backward` in `src/snapshot.rs`.

Looking at the code on v0.21.0 (and `main` is identical — no changes to `snapshot.rs` since the tag), the two functions are mutually tail-recursive. Each user key whose latest visible version is a tombstone — or where no version is visible to the snapshot — adds 2 stack frames instead of looping. So a backward range scan over a long run of tombstoned keys eats stack until it dies. ~500 consecutive tombstoned keys × 2 frames lines up with the 1001 levels in my crash report.

Reproducer:

```rust
const N: usize = 2000;
let (store, _temp_dir) = create_store();

{
    let mut tx = store.begin().unwrap();
    for i in 0..N {
        tx.set(format!("k{:08}", i).as_bytes(), b"v").unwrap();
    }
    tx.commit().await.unwrap();
}

// delete them all so the latest visible version of each is a tombstone
{
    let mut tx = store.begin().unwrap();
    for i in 0..N {
        tx.delete(format!("k{:08}", i).as_bytes()).unwrap();
    }
    tx.commit().await.unwrap();
}

let tx = store.begin().unwrap();
let snap = tx.snapshot.as_ref().unwrap();
let mut iter = snap.range(None, None).unwrap();
collect_snapshot_reverse(&mut iter).unwrap(); // boom
```

Output:

```
thread '...' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

With Rust's default test-thread stack the threshold is much lower than 2000 — a few hundred keys is enough. macOS gives bigger stacks at runtime, but you still get there with enough churn.

The forward path doesn't have this problem — `skip_to_valid_forward` is already a `loop`. Only the backward direction recurses.

Pretty similar in shape to #219 (the bplustree stack overflow), just in a different spot.

Opening a PR with the loop conversion and a regression test.

emmanuelm41:
Fixes #379.

`skip_to_valid_backward` and `find_latest_visible_backward` in `src/snapshot.rs` are mutually tail-recursive. Every user key whose latest visible version is a tombstone — or where no version is visible to the snapshot — adds 2 stack frames instead of looping. A backward range scan over a long run of such keys (which happens naturally after deletes) eats stack until it dies. We hit it on a real on-disk store, see the issue for the crash details.

The fix collapses the mutual recursion into a single `loop` inside `find_latest_visible_backward`. The two recursive tail calls become `continue`. `skip_to_valid_backward` becomes a one-line wrapper, kept for symmetry with `skip_to_valid_forward` (which was already a loop and is unaffected). Behavior is otherwise identical — same buffered-key handoff between iterations, same visibility rules, same outcomes.

Same shape of fix as #219, just in a different file.

## Test plan

- [x] New `test_backward_iter_does_not_stack_overflow_on_tombstones` in `src/test/snapshot_tests.rs`. Inserts 2000 keys, deletes them all, then reverse-scans the range. On `main` it aborts with `stack overflow`; with this patch it returns 0 entries in ~70ms.
- [x] `cargo test --lib` — 977 passed, 0 failed.
- [x] `cargo clippy --lib --no-deps -- -D warnings` clean.

### DIFF
--- a/src/snapshot.rs
+++ b/src/snapshot.rs
@@ -975,94 +975,103 @@ impl SnapshotIterator<'_> {
 	}
 
 	/// Skip to the next valid entry in backward direction.
-	/// More complex because we see oldest version first, need to find latest visible.
+	/// Thin wrapper kept for symmetry with `skip_to_valid_forward`; all real
+	/// work happens in `find_latest_visible_backward`, which loops internally.
 	fn skip_to_valid_backward(&mut self) -> Result<bool> {
-		// First check if we have a buffered entry from previous iteration
-		if self.has_buffered_back {
-			self.has_buffered_back = false;
-			// The buffered entry is already the start of a new user key
-			// We need to find the latest visible version of this key
-			return self.find_latest_visible_backward();
-		}
-
-		if self.merge_iter.valid() {
-			return self.find_latest_visible_backward();
-		}
-		self.has_current_back = false;
-		Ok(false)
+		self.find_latest_visible_backward()
 	}
 
-	/// Find the latest visible version of the current user key going backward.
-	/// Backward iteration sees oldest version first (lowest seq_num).
+	/// Find the latest visible version of the next user key going backward.
+	///
+	/// Backward iteration sees the oldest version of each user key first
+	/// (lowest seq_num) and must walk back to the newest, picking the latest
+	/// version that is visible to this snapshot. If that latest version turns
+	/// out to be a tombstone (or no visible version exists at all), we must
+	/// skip the user key entirely and examine the next one back.
+	///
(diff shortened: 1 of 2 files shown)


arriqaaq (approved):
Thank you for this @emmanuelm41

PR #396, opened 2026-08-17
fix(iter): correct range-iterator boundary handling on direction changes

### LINKED ISSUE #395: Range iterator: three boundary defects in direction handling (seek_last recovery, missing invalidation, lost committed key)
### Summary

Three defects in transaction range-iterator boundary handling, all deterministic with
minimal reproductions. Two root causes: a sticky range sentinel in the memtable
skiplist iterator, and a direction-change fixup block that resets the write-set cursor
instead of advancing it. The second exists as **four byte-identical copies** — one pair
in `TransactionRangeIterator`, one in `TransactionHistoryIterator`.

Affects `0.21.3`. Found by property-testing the iterator against a `BTreeMap` reference
model.

### Defect 1 — `seek_last()` cannot recover after `next()` passed the upper bound

```rust
txn.set(b"k08", b"v"); txn.set(b"k11", b"v");   // k11 is OUTSIDE the range below
txn.commit().await;
let mut iter = txn.range(b"k00".to_vec(), b"k09".to_vec())?;
iter.seek_first();   // -> k08, correct
iter.next();         // correctly invalidates
iter.seek_last();    // BUG: fails to reposition onto k08
```

**Root cause:** `SkiplistIterator::last()` walks back from `tail` guarded by
`is_valid()`, which returns false when the cursor sits on `upper_node` — a memoised
out-of-range sentinel set by `advance()` and never cleared. The loop therefore never
executes and the iterator stays pinned on the poisoned node.

The out-of-range key is required to reproduce, which is what makes this easy to miss.

**There is a live production path beyond the direct API.**
`MergingIterator::switch_to_backward` does
`if !child.iter.seek(target)? { child.iter.seek_last()?; }`. For a memtable child the
failed `seek` is precisely what memoises the sentinel, and the very next call is the
`seek_last()` that it pins. That is a route by which committed memtable rows can go
missing from a backward scan.

### Defect 2 — `next()`/`prev()` fail to invalidate, and `prev()` can lose a committed key

```rust
// (a) next() fails to invalidate
txn.set(b"k08", b"v"); txn.set(b"k09", b"v");   // uncommitted, >=2 entries
let mut iter = txn.range(b"k00".to_vec(), b"k10".to_vec())?;
iter.seek_last();    // -> k09, correct
iter.next();         // BUG: returns Ok(true), still valid

// (b) prev() fails to invalidate
txn.set(b"k01", b"v"); txn.set(b"k02", b"v");   // uncommitted, >=2 entries
let mut iter = txn.range(b"k00".to_vec(), b"k03".to_vec())?;
iter.seek_first();   // -> k01, correct
iter.prev();         // BUG: returns Ok(true), still valid

// (c) prev() silently LOSES a committed key
// commit k01, then uncommitted set k02, range [k00, k03)
iter.seek_first(); iter.next(); iter.prev();    // BUG: Ok(false), k01 never returned
```

**Root cause:** the direction-change fixup, present in four identical copies:

```rust
if !self.snapshot_iter.valid() || !self.ws_valid() {
    self.seek_ws_first();          // or seek_ws_last() in the mirror
} else if self.current_source == CurrentSource::Snapshot {
    self.advance_ws();
} else {
    self.snapshot_iter.next()?;
}
```

For an uncommitted transaction the snapshot iterator is always invalid, so the first
branch fires every time and **teleports** the write-set cursor back to the start/end
rather than advancing it. `advance_ws()` then walks it back and re-emits an
already-emitted key.

Case (c) is a different fault in the same block: it never repositions an *exhausted*
snapshot cursor, so a committed key that should be revisited on the way back is
dropped.

Worth noting explicitly, because it misled us twice: (a) is **not** a direction-switch
bug — `seek_last()` is the first positioning call, and there is no direction change.
The trigger is write-set cardinality >= 2. It is also not about tombstones; it
reproduces with two plain `set()` calls.

### Defect 3 — limited history iterators terminate one step early

The same block in `TransactionHistoryIterator`. With `HistoryOptions::with_limit()`
and a write-set entry present at a direction reversal, `prev()` returns `false` one
step before it should. `HistoryIterator::valid()` returns false when `limit_reached`,
and the block cannot distinguish that policy stop from genuine exhaustion.

### Fix

PR to follow. All three are fixed, the four copies are collapsed into one shared
helper, and every symptom above has a regression test committed failing first.

PeterKnego:
Fixes #395.

Three defects, two root causes, plus one latent issue the de-duplication surfaced.

### 1. Sticky range sentinel in the skiplist iterator

`SkiplistIterator::last()` walked back from `tail` guarded by `is_valid()`, which is
false when the cursor sits on the memoised `upper_node` — exactly the node the
backward walk must step off. Re-guarded on "is a real node".

The sentinels stay sticky, which is sound and worth stating since it looks suspicious:
nodes are immutable and never unlinked, `lower`/`upper` are written once at iterator
construction with no re-bounding API, and the bump arena never frees or recycles an
offset. So "this node is out of range" is a permanent fact, not a snapshot in time.
`last()` was the only reader that had to *start* on a sentinel; the other bound checks
are guarded by `is_valid()` too, but cannot land on the opposing sentinel.

### 2. Direction-change fixup reset the write-set cursor

The fixup existed in four byte-identical copies (`TransactionRangeIterator::next`/
`prev`, `TransactionHistoryIterator::next`/`prev`) and reset the write-set cursor
whenever the other source was invalid — which, for an uncommitted transaction, is
always. Replaced all four with one shared `reanchor_other_source()`: step the
non-current cursor one entry in the new direction, and re-anchor it only if genuinely
exhausted. That fixes both the re-emission and the case where an exhausted snapshot
cursor was never repositioned, silently dropping a committed key.

### 3. Limit budget across a re-anchor

De-duplicating surfaced that `HistoryIterator::valid()` is false on `limit_reached`,
which is a policy stop rather than exhaustion, and that `seek_first`/`seek_last` clear
the counter via `reset_all_state()`.

A boolean predicate cannot express the right behaviour here: given only
`seek_first`/`seek_last` as re-anchor primitives, it can choose to refund the whole
budget or forfeit it, and both are wrong. So this adds
`LSMIterator::reanchor_first`/`reanchor_last`, defaulting to `seek_first`/`seek_last`
(unchanged for the eleven position-only implementors), with the contract "reposition
without starting over". `HistoryIterator` overrides them to reposition while preserving
`entries_returned`, so a reverse pass runs on the remaining budget and the existing
truncation applies it.

`seek_first`/`seek_last` and the re-anchors now share one body so the pair cannot
drift.

### Also

`next()`/`prev()` on an already-invalid iterator now stay invalid and mutate nothing.
This matches every other `LSMIterator` in the crate and RocksDB's contract, and closes
a path where the old fixup could mutate state and let an exhausted iterator report
valid again. Recovery is what `seek_first`/`seek_last`/`seek` are for.

### Tests

Every symptom has a regression test, each committed failing first:

- `src/test/iterator_semantics_tests.rs` (10 tests) — one per symptom, plus four that
  pin the surrounding semantics, plus a post-flush variant confirming the SSTable/B+tree
  path does **not** share defect 1 (it re-seeks from scratch).
- `src/test/version_iterator_tests.rs` (3 new) — limit preserved across a direction
  switch, limit preserved when the data exhausts before the limit is reached, and one
  that pins the exact arithmetic: 3 versions with `limit(5)` must total exactly 5,
  which distinguishes the correct behaviour from refunding (6) and forfeiting (3).

`cargo test --lib`: 1040 passed, 0 failed, 2 ignored (both pre-existing and unrelated).
`cargo clippy --all --all-targets` and `cargo +nightly fmt --all -- --check` clean.

### Note for reviewers

These were found by a model-based property test that compares the iterator against a
`BTreeMap` reference model over randomised bidirectional walks. That harness is not in
this PR — it is a larger piece of test infrastructure — but I am happy to open it
separately if it would be useful. It is what turned "backward scans sometimes look
wrong" into the four minimal reproductions above.

### DIFF
--- a/src/memtable/skiplist.rs
+++ b/src/memtable/skiplist.rs
@@ -667,10 +667,20 @@ impl Skiplist {
 pub(crate) struct SkiplistIterator<'a> {
 	list: &'a Skiplist,
 	nd: *mut Node,
-	lower: Option<Vec<u8>>,   // Inclusive lower bound
-	upper: Option<Vec<u8>>,   // Exclusive upper bound
-	lower_node: *mut Node,    // Cached node at lower bound
-	upper_node: *mut Node,    // Cached node at upper bound
+	lower: Option<Vec<u8>>, // Inclusive lower bound
+	upper: Option<Vec<u8>>, // Exclusive upper bound
+	// `lower_node` / `upper_node` memoise the first node the cursor was ever
+	// found to sit outside the bounds on, so subsequent steps onto the same node
+	// can be rejected by pointer identity instead of a key comparison.
+	//
+	// They are deliberately sticky: skiplist nodes are immutable and never
+	// unlinked, and the bounds are fixed for the lifetime of the iterator, so
+	// "this node's key is < `lower`" / "this node's key is >= `upper`" are facts
+	// that stay true across re-seeks. What must NOT be assumed is that a cursor
+	// sitting on one of them is a dead end: `last()` has to *start* on an
+	// out-of-bounds node and walk back off it. See the note in `last()`.
+	lower_node: *mut Node, // Cached node known to be below the inclusive lower bound
+	upper_node: *mut Node, // Cached node known to be at or past the exclusive upper bound
 	encoded_key_buf: Vec<u8>, // Buffer for encoded key to return InternalKeyRef
 }
 
@@ -689,6 +699,18 @@ impl<'a> SkiplistIterator<'a> {
 	#[inline]
 	pub fn key_bytes(&self) -> &[u8] {
 		debug_assert!(self.is_valid());
+		self.node_key_bytes()
+	}
+
+	/// Key bytes of whatever node the cursor sits on, without `is_valid()`'s
+	/// in-bounds requirement.
+	///
+	/// Only for the internal bound walks, which legitimately inspect nodes that
+	/// are outside the iterator's range. The caller must guarantee `nd` is a real
+	/// node (neither sentinel, non-null).
+	#[inline]
+	fn node_key_bytes(&self) -> &[u8] {
+		debug_assert!(!self.nd.is_null() && self.nd != self.list.head && self.nd != self.list.tail);
 		unsafe { (*self.nd).get_key_bytes(&self.list.arena) }
 	}
 
@@ -752,10 +774,23 @@ impl<'a> SkiplistIterator<'a> {
 		if self.nd == self.list.head || self.nd == self.lower_node {
 			return;
 		}
-		// Check upper bound first - if entry is at or past upper, move backward
+		// Check upper bound first - if entry is at or past upper, move backward.
+		//
+		// This walk must NOT be guarded by `is_valid()`. `is_valid()` is false
+		// both for the list sentinels AND for a cursor parked on the memoised
+		// `upper_node`/`lower_node` — and `upper_node` is precisely the node this
+		// loop exists to step back off. Guarding on `is_valid()` made the loop
+		// body unreachable whenever a previous forward `advance()` had already
+		// memoised the list's physically last node as `upper_node`, leaving
+		// `last()` pinned on that out-of-range node and reporting the iterator
+		// invalid (the `seek_last()`-after-`next()`-ran-past-the-end bug).
+		//
+		// `lower_node` still terminates the walk below, because a node known to
+		// be below the inclusive lower bound means everything from here down is
+		// out of range.
 		if let Some(upper) = self.upper.as_deref() {
-			while self.is_valid() {
-				let key = self.key_bytes();
+			while self.nd != self.list.head && self.nd != self.list.tail && !self.nd.is_null() {
+				let key = self.node_key_bytes();
 				if (self.list.cmp)(upper, key) == Ordering::Greater {
 					// key < upper, so this entry is valid
 					break;
--- a/src/memtable/skiplist.rs
+++ b/src/memtable/skiplist.rs
@@ -676,9 +676,7 @@ pub(crate) struct SkiplistIterator<'a> {
 	// They are deliberately sticky: skiplist nodes are immutable and never
 	// unlinked, and the bounds are fixed for the lifetime of the iterator, so
 	// "this node's key is < `lower`" / "this node's key is >= `upper`" are facts
-	// that stay true across re-seeks. What must NOT be assumed is that a cursor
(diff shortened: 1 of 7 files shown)


src/memtable/skiplist.rs
```diff
@@ -752,10 +774,23 @@ impl<'a> SkiplistIterator<'a> {
 		if self.nd == self.list.head || self.nd == self.lower_node {
 			return;
 		}
-		// Check upper bound first - if entry is at or past upper, move backward
+		// Check upper bound first - if entry is at or past upper, move backward.
+		//
```
arriqaaq, line 776: Possible we remove these comments which track historical changes if is_valid is not being used?

src/memtable/skiplist.rs
```diff
@@ -667,10 +667,20 @@ impl Skiplist {
 pub(crate) struct SkiplistIterator<'a> {
 	list: &'a Skiplist,
 	nd: *mut Node,
-	lower: Option<Vec<u8>>,   // Inclusive lower bound
-	upper: Option<Vec<u8>>,   // Exclusive upper bound
-	lower_node: *mut Node,    // Cached node at lower bound
-	upper_node: *mut Node,    // Cached node at upper bound
+	lower: Option<Vec<u8>>, // Inclusive lower bound
+	upper: Option<Vec<u8>>, // Exclusive upper bound
+	// `lower_node` / `upper_node` memoise the first node the cursor was ever
```
arriqaaq, line 672: I don't think cursor related info should be in the comment for definition of a filed

src/memtable/skiplist.rs
PeterKnego: Done in 89ceb05 — the block is down from 12 lines to 5.

I kept a compressed statement of the invariant rather than dropping the comment outright: `is_valid()` returns false when `nd == upper_node`, so guarding this walk on it skips exactly the case the walk exists to handle. Without a note there it reads like a missing guard and is an easy thing to "fix" back. The bug narration and the `advance()` backstory are gone.

src/memtable/skiplist.rs
PeterKnego: Moved in 89ceb05. The field comment now only describes what the two fields hold and why they are sticky; the sentence about `last()` starting on an out-of-bounds node and walking back off it is gone from the declaration and lives at the use site in `last()` instead.

PR #401, opened 2026-08-25
Add get_for_update locked point reads to transactions

tobiemh:
## What

Adds `Transaction::get_for_update`: a locked point read with the same read semantics as `get` (read-your-own-writes included) that registers the key for commit-time conflict detection. A transaction that locked a key commits only if no other transaction committed a write to that key after its snapshot.

## How

- New `read_set: BTreeMap<Key, u32>` on `Transaction` (key → savepoint number at first registration), populated by `get_for_update`. Requires ReadWrite mode (`TransactionReadOnly` otherwise, since validation cannot run at commit).
- At commit, the keys passed to the oracle conflict check are the union of write-set keys and read-set keys; the write payload (publish/WAL/apply) remains exactly the write set — locked reads write nothing. Read-set keys shadowed by a write are dropped before the check since the write key covers them.
- A commit with an empty write set but non-empty read set runs a new `CommitPipeline::check_conflicts`: after the same shutdown and background-error gates as a normal commit, it acquires `write_mutex` and runs `oracle.check` under it — the same lock under which every normal commit runs check + seq-alloc + publish + WAL — so validation is atomic with respect to concurrent committers. It deliberately skips the write-stall gate and commit semaphore because it never enqueues, allocates no seq, and writes nothing.
- `rollback()` clears the read set; `rollback_to_savepoint()` retains entries by savepoint number, following the write-set pattern.
- Conflicts surface as the existing `TransactionWriteConflict` (same oracle mechanism, no new downstream mapping needed); `TransactionRetry` can surface from the same check exactly as for write commits.

## Tests

10 new transaction tests (conflict, control, locked-read-only conflict and clean commit, value semantics, plain-get non-registration, read-only rejection, savepoint interactions) plus an oracle concurrency test racing a check-only commit against a writer. Full suite passes.

## Motivation

Consumed by SurrealDB's `SELECT ... FOR UPDATE` (surrealdb PR #998) as this engine's conflict-registration primitive.

### DIFF
--- a/src/commit.rs
+++ b/src/commit.rs
@@ -10,6 +10,7 @@ use crate::batch::Batch;
 use crate::error::{Error, Result};
 use crate::oracle::CommitOracle;
 use crate::stall::WriteStallController;
+use crate::Key;
 
 const MAX_CONCURRENT_COMMITS: usize = 8;
 const DEQUEUE_BITS: u32 = 32;
@@ -250,7 +251,13 @@ impl CommitPipeline {
 		self.oracle.reset_for_restore(max_seq);
 	}
 
-	pub(crate) async fn commit(&self, mut batch: Batch, sync: bool, start_seq: u64) -> Result<()> {
+	pub(crate) async fn commit(
+		&self,
+		mut batch: Batch,
+		sync: bool,
+		start_seq: u64,
+		read_set: &[Key],
+	) -> Result<()> {
 		if self.shutdown.load(Ordering::Acquire) {
 			return Err(Error::PipelineStall);
 		}
@@ -290,15 +297,26 @@ impl CommitPipeline {
 		//      enqueued. Roll back those entries (seq-match guard preserves concurrent
 		//      overwriters), drain the queue slot, release the lock, and return the error.
 		//
-		// Keys are derived from `batch.entries` (single source of truth).
-		// Duplicate keys within a batch (e.g. from savepoint history) are
-		// harmless: oracle.check/publish are idempotent on the same key.
+		// Write keys are derived from `batch.entries` (single source of
+		// truth); `read_set` contributes check-only keys that are validated
+		// but never published or written. Duplicate keys within a batch
+		// (e.g. from savepoint history) are harmless: oracle.check/publish
+		// are idempotent on the same key.
 		let (processed_batch, allocated_seq): (Batch, u64) = {
 			let _guard = self.write_mutex.lock();
 
-			// Validate against the oracle. No state has changed yet; on
-			// failure `?` simply returns the error to the caller.
-			self.oracle.check(batch.entries.iter().map(|e| e.key.as_slice()), start_seq)?;
+			// Validate against the oracle. The checked keys are the union of
+			// the batch's write keys and the caller's locked-read keys; only
+			// the write keys are published below. No state has changed yet;
+			// on failure `?` simply returns the error to the caller.
+			self.oracle.check(
+				batch
+					.entries
+					.iter()
+					.map(|e| e.key.as_slice())
+					.chain(read_set.iter().map(|k| k.as_slice())),
+				start_seq,
+			)?;
 
 			let count = batch.count() as u64;
 			let seq_num = self.log_seq_num.fetch_add(count, Ordering::SeqCst);
@@ -401,6 +419,28 @@ impl CommitPipeline {
 		}
 
 		complete_rx.await.map_err(|_| Error::PipelineStall)?
+	}
+
+	/// Validate `keys` against the commit oracle without allocating a seq,
+	/// publishing oracle entries, or writing anything.
+	///
+	/// This is the commit path for a transaction whose only commit-time
+	/// obligation is conflict validation of locked reads. The check runs
+	/// under `write_mutex`, which makes it atomic with respect to concurrent
+	/// committers: every commit that entered the critical section before us
+	/// has already published its oracle entries and is visible to this
+	/// check, and every commit that enters after us serializes after this
+	/// transaction's validation point.
+	pub(crate) fn check_conflicts(&self, keys: &[Key], start_seq: u64) -> Result<()> {
+		if self.shutdown.load(Ordering::Acquire) {
+			return Err(Error::PipelineStall);
+		}
+
+		// Check for background errors before proceeding
+		self.env.check_background_error()?;
+
+		let _guard = self.write_mutex.lock();
+		self.oracle.check(keys.iter().map(|k| k.as_slice()), start_seq)
 	}
 
 	#[cfg(test)]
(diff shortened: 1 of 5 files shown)


src/transaction.rs
```diff
@@ -743,8 +788,13 @@ impl Transaction {
 			return Err(Error::TransactionReadOnly);
 		}
 
-		// If there are no pending writes, there's nothing to commit, so return early.
+		// If there are no pending writes, only the locked reads (if any) need
+		// commit-time validation; nothing is persisted.
 		if self.write_set.is_empty() {
+			if !self.read_set.is_empty() {
+				let read_set: Vec<Key> = std::mem::take(&mut self.read_set).into_keys().collect();
```
tobiemh: **`read_set` is drained before the fallible check, so a retried `commit()` silently reports success.**

`std::mem::take` empties the read set *before* `check_conflicts` can fail, and the `?` error path never sets `closed`. The transaction is left open with no locks recorded, so the conflict is not reproducible on a second attempt:

```rust
let mut txn1 = store.begin().unwrap();
txn1.get_for_update(&key).unwrap();          // read_set = {k}

let mut txn2 = store.begin().unwrap();
txn2.set(&key, b"v2").unwrap();
txn2.commit().await.unwrap();

txn1.commit().await;  // Err(TransactionWriteConflict)  — read_set now empty
txn1.commit().await;  // Ok(())   <-- conflicted txn reports success
```

I confirmed this with a probe test: the second `commit()` returns `Ok(())`.

The same drain-before-fail shape exists on the write path at line 815 (`std::mem::take(&mut self.read_set)` before `core.commit(...).await?`). There the lost writes on retry are pre-existing, but the *silently dropped locked-read validation* is new.

Suggested fix: take the sets only after the fallible step succeeds, or mark the transaction poisoned/closed on the error path so a retry can't report `Ok`.

src/transaction.rs
```diff
@@ -413,6 +423,41 @@ impl Transaction {
...
+	/// check at commit: if another transaction commits a write to the key
+	/// after this transaction started, the commit fails with
+	/// [`Error::TransactionWriteConflict`]. The locked read itself persists
+	/// nothing.
+	///
+	/// Requires a `ReadWrite` transaction: read-only transactions cannot run
+	/// commit-time validation, and write-only transactions cannot read.
+	///
+	/// [`get`]: Transaction::get
+	pub fn get_for_update<K>(&mut self, key: K) -> Result<Option<Value>>
```
tobiemh, line 451: **Two transactions that both lock the same key don't conflict with each other — write skew survives the lock.**

Locked-read keys join `oracle.check` but are never `oracle.publish`ed, so a locked read is invisible to other transactions. The classic read-modify-write case is covered (the write key carries the conflict), but the lock-then-write-elsewhere case is not:

```rust
let mut t1 = store.begin().unwrap();
let mut t2 = store.begin().unwrap();

t1.get_for_update(&shared).unwrap();   // both lock the same key
t2.get_for_update(&shared).unwrap();

t1.set(b"out1", b"a").unwrap();        // decisions derived from `shared`
t2.set(b"out2", b"b").unwrap();

t1.commit().await;  // Ok(())
t2.commit().await;  // Ok(())   <-- both lockers commit
```

I confirmed this with a probe test: `r1=Ok(()) r2=Ok(())`.

Preventing this would need locked reads to be visible to other lockers (a separate holder map), not just publishing them to the oracle — publishing would make readers abort writers, which is worse. So this may well be the intended scope, but since the stated consumer is `SELECT ... FOR UPDATE`, the doc comment should say so explicitly: the guarantee is *"no other transaction committed a write to this key after my snapshot"*, **not** mutual exclusion between lockers. Otherwise callers will reasonably assume `FOR UPDATE` serializes concurrent lockers and reach for it to prevent exactly this write skew.

src/commit.rs
```diff
@@ -250,7 +251,13 @@ impl CommitPipeline {
 		self.oracle.reset_for_restore(max_seq);
 	}
 
-	pub(crate) async fn commit(&self, mut batch: Batch, sync: bool, start_seq: u64) -> Result<()> {
+	pub(crate) async fn commit(
+		&self,
+		mut batch: Batch,
+		sync: bool,
+		start_seq: u64,
+		read_set: &[Key],
```
tobiemh, line 259: **`read_set` is accepted here but silently ignored when the batch is empty.**

A few lines below (line 268) there's an early return that predates this parameter:

```rust
if batch.is_empty() {
    return Ok(());
}
```

That runs *before* `oracle.check`, so an empty batch with a non-empty `read_set` returns `Ok(())` with zero conflict validation.

Not reachable from today's only caller — `Transaction::commit` reaches the write path only when `write_set` is non-empty, and every `write_set` key contributes at least one `batch.entries` record — but the new signature makes it an easy trap for a future caller, and the failure is silent (an unvalidated locked read reported as a clean commit) rather than loud.

Either move the check above the early return, or make the contract explicit at the top of the function:

```rust
debug_assert!(
    read_set.is_empty() || !batch.is_empty(),
    "read_set would be silently dropped by the empty-batch early return",
);
```

src/transaction.rs
tobiemh: Fixed in c6af669 — commit is now terminal: the fallible work moved into a private `commit_inner`, and the outer `commit` closes the transaction on success and failure alike, so a retry gets `TransactionClosed` instead of a false `Ok`. The same fix covers the write path, with regression tests for both.

src/transaction.rs
tobiemh: Documented in c6af669, behaviour unchanged: the guarantee is snapshot validation, not mutual exclusion — locked reads publish nothing, so concurrent lockers of the same key both commit, and callers needing lockers to conflict must write the key. Pinned with `concurrent_lockers_of_same_key_both_commit`.

src/commit.rs
tobiemh: Fixed in c6af669 — the empty-batch path now returns early only when the read set is also empty, and otherwise runs `check_conflicts` under `write_mutex`. Covered by `test_empty_batch_validates_read_set`.

arriqaaq:
Do we call this Serializable and then differentiate from SI? Currently we are proper snapshot isolation, which doesn't really track read set, and I don't want to change that if we ever need to switch back to it. I can do that if you are ok in the same PR

tobiemh:
This PR deliberately leaves SI untouched: plain `get`s still track nothing, and a transaction that never calls `get_for_update` behaves exactly as before — the read set only exists for keys explicitly registered through `get_for_update`. So there's nothing to switch back from.

I'd also avoid calling it Serializable, because it's weaker than that by design: per-key conflict registration (the same shape as RocksDB's `GetForUpdate` and TiKV's `lock_keys`), with no automatic read tracking, no phantom protection, and no locker-vs-locker conflicts. It backs `SELECT ... FOR UPDATE` in SurrealDB, which needs exactly "validate these specific keys at commit" — a transaction-wide mode would widen the abort surface of every plain read, which is what we want to avoid.

A real `Serializable` isolation level that tracks the full read set would be a great addition, but I'd take that as a follow-up rather than this PR, so we can land the primitive now.

arriqaaq (approved):

arriqaaq (approved):

PR #403, opened 2026-09-19
Treat missing ReadOptions bounds as unbounded in range_with_options

sawy3r:
`ReadOptions` documents a `None` bound as unbounded, but `Transaction::range_with_options` turned each missing bound into the empty key with `unwrap_or_default()` before building the iterator. That had two effects:

- With no upper bound, the upper bound became the empty key, so the scan returned no rows at all.
- With a lower bound set and no upper bound, the write-set lookup became `BTreeMap::range(lower..vec![])`, which panics because start is greater than end.

`Snapshot::range` already takes `Option<&[u8]>` bounds, so the fix passes the options through unchanged and builds the write-set range from `Bound`s instead of a `Range`. `Transaction::range` always supplies both bounds, so its behaviour is unchanged.

The new test covers lower-only, upper-only and fully unbounded scans, forward and reverse, including a key written in the same transaction beyond the committed data. On current main it panics.


arriqaaq:
Thank you @sawy3r

arriqaaq (dismissed):

arriqaaq:
Could you fix the format @sawy3r? And check the CI, will merge post that

sawy3r:
Thanks @arriqaaq, pushed a commit for the format job now.

arriqaaq (approved):

